PR #16, opened 2025-06-10
[BACKEND] Lowering i1 ptr to the PtrDialect on aipu

zhzhcookie:


third_party/aipu/backend/compiler.py
```diff
@@ -100,6 +100,7 @@ def make_linalg(mod, metadata, opt):
         pm.enable_debug()
         # add pass here
         aipu.passes.convert.add_triton_to_linalg_pipeline(pm)
+        aipu.passes.convert.add_triton_to_ptr(pm)
```
Lucien0, line 103: Seems already in TritonToLinalgExperimentalPass, why here need do it again?

PR #24, opened 2025-06-18
[BACKEND] Add fmod & pow in aipu libdevice

ikushare:
The extension of the external math library math.fmod requires adding definitions in libdevice.py of AIPU


Lucien0 (approved):

Lucien0:
LGTM

PR #54, opened 2025-07-24
[BACKEND] Support cambricon

chenmiao1919:
[BACKEND] Support cambricon


python/setup.py
```diff
@@ -594,29 +596,72 @@ def get_install_requires():
     return install_requires
 
 
+packages = get_packages()
+package_dir = helper.CommonUtils.get_package_dir(get_packages())
+
+if helper.flagtree_backend == "cambricon":
```
Galaxy1458, line 612: This part of the changes belongs to `cambricon` specialization changes. We hope to place these specialization functions in `python/setup_tools/utils/cambricon.py` and call them in `setup.py`.

chenmiao1919: done

Galaxy1458 (commented):
LGTM

Galaxy1458 (approved):

PR #55, opened 2025-07-24
Iluvatar code decoupling

toudefu:
iluvatar python code decouple 0724


python/setup.py
```diff
@@ -654,4 +652,11 @@ def get_install_requires():
             "tabulate",
         ],
     },
-)
+}
+
+# 根据环境变量决定是否添加 package_dir
```
sgjzfzzf: Please replace them with English comments.

python/triton/backend_loader.py
```diff
@@ -0,0 +1,13 @@
+import importlib
+from triton.backend_context import get_backend
+
+def load_backend(platform: str):
+    try:
```
sgjzfzzf: I didn't find any references to this part in the code. If we don't need this module, why do we need to import them here? Just for runtime backend support check?

python/triton/compiler/code_generator.py
```diff
@@ -1261,6 +1261,11 @@ def ret(self, node: ast.Call):
...
 
 def kernel_suffix(signature, specialization):
+    from triton.backend_context import get_backend
+    _backend = get_backend()
+    if _backend and hasattr(_backend, "kernel_suffix"):
+        return _backend.kernel_suffix(signature, specialization)
+
     # suffix format:
     # <argid><'c' if equal to 1><'d' if divisible by 16><'e' if divisible by 8>
     suffix = ''
```
sgjzfzzf: Such design looks a little strange...... Can we rewrite them as a default implementation for `Backend`, and ask the specific backend implementation to override it? Will this design introduce any more problem?

PR #95, opened 2025-09-15
[Docker] Upcase "AS" to solve warnings by docker

sgjzfzzf:
[OTHER] Docker will warn if the case of FROM and AS are not aligned. Fix it.


zhzhcookie (approved):
LGTM

PR #99, opened 2025-09-17
[Common] refactor UnifiedHardware

sgjzfzzf:
[BACKEND] This PR refactored `UnifiedHardware`. We define a base class and provides its default method implementations. When a new hardware is included, it needs to inherit the base class and override the methods if needed. Then they need to provide their own `createUnifiedHardwareManager` function to create it. We put a default implementation but mark it as "weak" so it could be overridden.


Galaxy1458 (approved):
LGTM

PR #123, opened 2025-10-15
[BACKEND] Fix the bug in editable mode and the issue where dynamic libraries cannot be automatically uninstalled.

SabYic:
Also add the change for CI to check for the use of triton-opt


Galaxy1458 (approved):
LGTM

PR #136, opened 2025-10-23
[FixBug]Fixed the bug where the submodule npu-ir could not be pulled synchronously when pulling the third-party library ascend

SabYic:


Galaxy1458 (approved):
LGTM

PR #143, opened 2025-11-06
[FixBug] TritonGPU: newExtOrBroadcastType assignment

BabarZKhan:


zhzhcookie (commented):
Thank you for your attention to FlagTree.
This typo will be fixed in commit 4b36a8a88af2b81ce2fb1d9245bd425268071338, so we will follow up with a fix in the corresponding triton version.

PR #165, opened 2025-12-02
[BACKEND] Add pass management interfaces for ascend

toudefu:


lib/flagtree/Common/UnifiedHardware.cc
```diff
@@ -15,6 +15,8 @@ int UnifiedHardware::getDMATag() const { return 0; }
 
 int UnifiedHardware::getSharedMemoryTag() const { return 0; }
 
+bool UnifiedHardware::getIncubatedTag() const { return 0; }
```
zhzhcookie: return false

toudefu: ok

PR #172, opened 2025-12-06
[BACKEND] Support sqmma&wmma feature on mthreads backend.

lingfengqiu:
Adapt flagtree with latest triton_musa (with sqmma and wmma feature) to support BAAI Q4 test.


zhzhcookie (approved):
LGTM

PR #204, opened 2025-12-19
[BACKEND] Iluvatar revert workaround in getPointer

Salamanca001:
Revert earlier workaround in jit.py, which could incur hang for some UT in flaggems.


Galaxy1458 (approved):
LGTM

zhzhcookie (approved):
LG

PR #221, opened 2025-12-23
[DOC] Update README for the project

tengqm:
This PR reworks the project README file for clarity.

Details about the build, install, testing instructions are extracted and spilled into the `docs/en` and `docs/cn` directories.


docs/cn/build.md
```diff
@@ -0,0 +1,344 @@
...
+- [triton\_v3.4.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.4.x)
+- [triton\_v3.5.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.5.x)
+
+各主干分支均为保护分支且地位相等。
+
+各后端完整构建命令如下：
+
+### 天数智芯
+
+[Iluvatar](https://github.com/flagos-ai/flagtree/tree/main/third_party/iluvatar/)
```
zhzhcookie: 与开发中使用的后端名 iluvatar 保持一致更便于用户设置环境变量、查找后端。

README.md
```diff
@@ -1,284 +1,38 @@
-<div align="right"><a href="/README_cn.md">中文版</a></div>
+<div align="right"><a href="./README_cn.md">中文版</a></div>
 
-## <img width="30" height="30" alt="FlagTree-GitHub" src="https://github.com/user-attachments/assets/d8d24c81-6f46-4adc-94e2-b89b03afcb43" /> FlagTree
+## Overview
```
zhzhcookie, line 4: 整体将首页说明整理得非常简洁，但不符合惯例。可参考 Triton、PyTorch、triton-cpu、facebookexperimental/triton、triton-shared、vllm、cub、cccl、Megatron-LM、cuda-tile ... 等与本仓库密切相关的原版仓库首页风格，不建议参考非本技术方向仓库风格，更不建议参考国内仓库的风格。

documents/zh-cn/install.md
```diff
@@ -0,0 +1,51 @@
...
+| 后端     | 安装命令                                                   | Triton 版本 | 支持的 Python 版本 |
+| -------- | ---------------------------------------------------------- | ----------- | ------------------ |
+| nvidia   | python3 -m pip install flagtree==0.3.0rc1 $RES             | 3.1         | 3.10, 3.11, 3.12   |
+| nvidia   | python3 -m pip install flagtree==0.3.0rc1+3.2 $RES         | 3.2         | 3.10, 3.11, 3.12   |
+| nvidia   | python3 -m pip install flagtree==0.3.0rc1+3.3 $RES         | 3.3         | 3.10, 3.11, 3.12   |
+| iluvatar | python3 -m pip install flagtree==0.3.0rc2+iluvatar3.1 $RES | 3.1         | 3.10               |
+| mthreads | python3 -m pip install flagtree==0.3.0rc3+mthreads3.1 $RES | 3.1         | 3.10               |
+| ascend   | python3 -m pip install flagtree==0.3.0rc1+ascend3.2 $RES   | 3.2         | 3.11               |
+| hcu      | python3 -m pip install flagtree==0.3.0rc2+hcu3.0 $RES      | 3.0         | 3.10               |
+| enflame  | python3 -m pip install flagtree==0.3.0rc1+enflame3.3 $RES  | 3.3         | 3.10               |
```
zhzhcookie, line 49: 表头不建议空格对齐，否则后续修改会导致该表头频繁编辑，不利于diff维护。且不同浏览器上呈现效果不一致，当前看来也不是对齐的。

documents/zh-cn/tests.md
```diff
@@ -0,0 +1,18 @@
+[English](../en/CONTRIBUTING.md)
+
+## 运行测试
+
+安装完成后可以在设备支持的环境下运行测试。对于 NVIDIA 后端，可以执行：
+
+```shell
+cd python/test
+python3 -m pytest -s
+```
```
zhzhcookie, line 10: 未更新至最新版本

documents/en/build.md
```diff
@@ -0,0 +1,319 @@
...
+- [triton\_v3.2.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.2.x) for Triton 3.2.x
+- [triton\_v3.3.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.3.x) for Triton 3.3.x
+- [triton\_v3.4.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.4.x) for Triton 3.4.x
+- [triton\_v3.5.x](https://github.com/flagos-ai/flagtree/tree/triton_v3.5.x) for Triton 3.5.x
+
+All these protected branches are maintained with equal foot.
+
+Complete build commands for each backend:
+
+### Illuvatar
```
zhzhcookie, line 17: 与开发中使用的后端名 iluvatar 保持一致更便于用户设置环境变量、查找后端。

tengqm: This should not be treated the same as backend (environment) name. It is simply the canonical name for the chip/model. For example, we can use 'nvidia' as the backend name, but when referring to the chip or model, the canonical name should be 'NVIDIA'.

documents/zh-cn/install.md
tengqm: This is for readability's sake. Without intra-cell padding, it is hard to find which Triton version is used by a specific hardware.
The width of the second column could be a problem, we can increase it if necessary.
That kind of an update would be done only when necessary.

README.md
tengqm: There is no so-called 'convention'.
The README file is not the home of all setup/install/changelog/promotion text.
We are aiming to make it simple and concise, ACROSS ALL FLAGOS projects.
For new users/contributors, they only want an overview of the project.
For users who come back to the project, they don't read the README at all. They want to check the specific information instead.

docs/cn/build.md
tengqm: This is not the name of an environment variable. No one would think that way. Right?

documents/zh-cn/tests.md
tengqm: Thanks. Lemme check.

tengqm: Done.

documents/zh-cn/install.md
zhzhcookie: > This is for readability's sake. Without intra-cell padding, it is hard to find which Triton version is used by a specific hardware. The width of the second column could be a problem, we can increase it if necessary. That kind of an update would be done only when necessary.

In fact, when I look at it in my browser, the table header isn't aligned either, so I think its readability is not as good as a compact arrangement.

docs/cn/build.md
zhzhcookie: > This is not the name of an environment variable. No one would think that way. Right?

In my opinion, using the same name for the same concept and different names for different concepts makes it less likely to confuse users, reducing the cost of explanation and the probability of misoperation.

documents/zh-cn/install.md
tengqm: Ah I see. It is not perferctly aligned in browser because the font substitution in the browser.
When editing the file in an editor (matters to maintainers) it is perfectly aligned.
When viewing the rendered result as a HTML table (matters to users), it is perfectly aligned.

documents/en/build.md
tengqm: This is exactly the thing we want to advocate across all projects. We are supposed to pay more attention to spelling, especially for words that a proper names. You can call the same backend using different variable names for convenience, but in user-facing documentation, we should always use the proper names rather than the (environment) variable names.

documents/zh-cn/install.md
zhzhcookie: The webpage rendering result was originally aligned as well. It is a feature provided by md.

documents/en/build.md
zhzhcookie: > This is exactly the thing we want to advocate across all projects. We are supposed to pay more attention to spelling, especially for words that a proper names. You can call the same backend using different variable names for convenience, but in user-facing documentation, we should always use the proper names rather than the (environment) variable names.

I have reservations about that. In fact, it is precisely because users made mistakes on this matter and came to ask us that we pay attention to consistency. Many interpretation costs can be completely eliminated at the source.

documents/zh-cn/install.md
zhzhcookie: Here are some examples of open source projects:

image

image

tengqm: Thanks for sharing, @zhzhcookie.
I think I know well how different camps think of the other side.
Based on the screenshots, I believe that properly align the markdown table cells makes the markdown source more readable. I am not saying that markdown table is good for cases where some cells have long texts or cases where the column widths have to be frequently adjusted. In those cases, I'd suggest embedded HTML tables. The rule of thumb is to make the source readable, maintainable (the precondition is that the rendered HTML is the same). I don't care what the rest of the world is doing. I only care what solution is best suited for the problem at hand.

documents/en/build.md
tengqm: Correct. We are trying our best to establish a standard here. For example, stop writing "arm" or "nvidia" in the documentation. It is as ridiculous as writing "Oracle" as "oracle" or "CPU" as "cpu". It is perfectly fine to use lower-case strings as variable names, but it is not okay to change a company name freely. It reads unprofessional.

This page we are talking about the backends we support. We use the names everyone can understand, i.e. the proper names. When we are talking about the symbol names we use to represent a chip/backend, when we are explaining how to configure a specific backend, we use the variable names.

tengqm:
@zhzhcookie As a compromise, I have revised the list format for backend, as exemplified below:

image

tengqm:
conflict with another PR which has been merged.

PR #223, opened 2025-12-24
[Ascend] triton_patch python update and decouple

liuyunqi20:


python/triton/language/core.py
```diff
@@ -1622,7 +1622,7 @@ def dot_scaled(lhs, lhs_scale, lhs_format, rhs, rhs_scale, rhs_format, acc=None,
```
zhzhcookie: \# flagtree backend specialization add new params: "care_padding"

python/triton/runtime/jit.py
zhzhcookie, line 838: \# flagtree backend specialization add new params: "options"

PR #248, opened 2025-12-29
[FEAT]enable use wheel as external LLVM

starrryz:
Resolves issue#211

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -326,7 +326,6 @@ def open_url(url):
     headers = {
         'User-Agent': user_agent,
     }
-    print("[DEBUG] downloading url =", repr(url))
     request = urllib.request.Request(url, None, headers)
     # Set timeout to 300 seconds to prevent the request from hanging forever.
     return urllib.request.urlopen(request, timeout=300)
--- a/setup.py
+++ b/setup.py
@@ -283,10 +283,10 @@ def get_llvm_package_info_legacy():
 
 
 def get_llvm_package_info():
-    LLVM_WHEEL_PKG = "llvm-wheel"  
+    LLVM_WHEEL_PKG = "llvm-wheel"
     llvm_installed = is_llvm_wheel_installed(LLVM_WHEEL_PKG)
     llvm_envs = has_llvm_env_vars()
-    # rule1 : if both call erroe
+    # rule1 : if both exist, fail
     if llvm_installed and llvm_envs and not os.environ.get("USE_LLVM_WHEEL_BUILD"):
         raise RuntimeError(
             "ERROR: LLVM wheel is installed, but LLVM-related environment variables are set:\n"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -415,6 +489,7 @@ def run(self):
             raise RuntimeError("CMake >= 3.20 is required")
 
         for ext in self.extensions:
+
```
sgjzfzzf: Remove unnecessary blank lines.

setup.py
```diff
@@ -247,6 +282,45 @@ def get_llvm_package_info():
     return Package("llvm", name, url, "LLVM_INCLUDE_DIRS", "LLVM_LIBRARY_DIR", "LLVM_SYSPATH", sym_name=sym_name)
 
 
+def get_llvm_package_info():
+    LLVM_WHEEL_PKG = "llvm-wheel"  
+    llvm_installed = is_llvm_wheel_installed(LLVM_WHEEL_PKG)
+    llvm_envs = has_llvm_env_vars()
+    # rule1 : if both call erroe
```
sgjzfzzf: Typo error. And I think "fails" might be better?

setup.py
```diff
@@ -247,6 +282,45 @@ def get_llvm_package_info():
...
+            "Please unset them to avoid conflicts."
+        )
+    # rule2：wheel installed & no env → use wheel
+    if llvm_installed:
+        include_dir, lib_dir, llvm_root = get_llvm_paths_from_wheel(LLVM_WHEEL_PKG)
+
+        print("[DECISION] Using LLVM from wheel package")
+        print("  include:", include_dir)
+        print("  lib:    ", lib_dir)
+        print("  root:   ", llvm_root)
```
sgjzfzzf: Delete unnecessary prints.

setup.py
```diff
@@ -247,6 +282,45 @@ def get_llvm_package_info():
...
+        
+        # env variables will not appear out of python process
+        os.environ["USE_LLVM_WHEEL_BUILD"] = "1"
+        os.environ["LLVM_SYSPATH"] = llvm_root
+        os.environ["LLVM_INCLUDE_DIRS"] = include_dir
+        os.environ["LLVM_LIBRARY_DIR"] = lib_dir
```
sgjzfzzf, line 303: Where does the construction happen? Do we need to unset them somewhere else? It's important to avoid polluting the environment of the whole process.

Evaluate it please(I'm also not sure on that🤔)

lib/Conversion/FlagTreeToLLVM/CMakeLists.txt
sgjzfzzf, line 7: This modification is merged. Remove it from this PR.

sgjzfzzf (commented):
Remember running `pre-commit install` before pushes. I haven't added the branch protection rules yet but it's still important to make our PRs clean.

setup.py
starrryz: this envs will become members of LLVM package info and will appear as attrs of CMake as -DLLVM_SYSPATH which is clean for building(instead of the initial method : llvm_root(the real path) become the attr of LLVM package). 
envs will be inherited by sub process, so it might have influence in the future. now it is rather safe.

sgjzfzzf (approved):

PR #251, opened 2025-12-29
[BACKEND] Solve addresspace issue

### LINKED ISSUE #209: [Feature Request] Construct `ExtractOp`s by Inputs' Memory Space
### 🚀 The feature, motivation and pitch

`kSharedAddressSpace` is defined as a constant 3 in `lib/Dialect/FlagTree/IR/Ops.cpp` to describe the shared memory address space. It may lead to some problems when more memory spaces occur.

Update the definitions of `ExtractOp`s in `include/triton/Dialect/FlagTree/IR/FlagTreeOps.td`, and delete the self-defined builders. The new generated builders may require the users to assign the result type, and then get the memory space by checking the llvm function arguments and pass them to `ExtractOp`s in `python/src/flagtree_ir.cc` to avoid the constant definitions.

### Alternatives

_No response_

### Additional context

_No response_

SabYic:
resolve #209


python/examples/flagtree/03-matrix-multiplication.py
sgjzfzzf, line 1: Also modify this in other examples.

python/src/flagtree_ir.cc
```diff
@@ -41,6 +44,8 @@ flagtree::DSLRegionOp FlagTreeOpBuilder::createEdslRegionByLLVMFunc(
   SmallVector<Type> operandTys = llvm::map_to_vector(
       operands, [](Value value) -> Type { return value.getType(); });
   IRMapping mapper;
+  auto ptrTy = dyn_cast<LLVM::LLVMPointerType>(func.getArgument(0).getType());
+  unsigned as = ptrTy.getAddressSpace();
```
sgjzfzzf: Use `uint32_t` to make it more readable.

python/examples/flagtree/01-vector-add.py
```diff
@@ -11,7 +11,7 @@
 
 
 @dialect(name="mlir")
-def edsl(output: InOut["?xf32"], x: Input["?xf32"], y: Input["?xf32"]):  # noqa: F722
+def edsl(output: InOut["memref<?xf32,3>"], x: Input["memref<?xf32,3>"], y: Input["memref<?xf32,3>"]):  # noqa: F722
```
sgjzfzzf: Add whitespaces between `,` and `3` for readability for all annotations, not only this example.

sgjzfzzf (approved):

PR #279, opened 2026-01-08
[HINT] remove debug info for shmem hints

i3wanna2:
remove debug info for shmem hints
debug features will be controlled by environment variables in a future update


zhzhcookie (approved):
LGTM

PR #297, opened 2026-01-20
[BUILD] Update the git path for ascendnpu-ir

CHAIYD-2025:
[BUILD]:This commit is to update the git URL for ascendnpu-ir.


Galaxy1458 (approved):
LGTM

PR #342, opened 2026-02-03
[Build] Add FlagtreeConfig Class for triton_v3.4.x

SabYic:


python/setup_tools/utils/aipu.py
```diff
@@ -1,3 +1,7 @@
 def precompile_hock(*args, **kargs):
     default_backends = kargs["default_backends"]
-    default_backends.append('flir')
+    default_backends_list = list(default_backends)
+    default_backends_list.append('flir')
```
sgjzfzzf: Replace it by
```suggestion
    default_backends_list = [*default_backends, "flir"]
```
Would this be more readable?

python/setup_tools/utils/aipu.py
```diff
@@ -1,3 +1,7 @@
...
+    kargs["default_backends"] = tuple(default_backends_list)
+    default_backends = tuple(default_backends_list)
+    return default_backends
```
sgjzfzzf: Append an "\n" to the file end.

python/setup_tools/utils/tools.py
```diff
@@ -10,12 +10,58 @@
...
+        "cambricon": "mlu",
+    }))
+
+    def __post_init__(self):
+        object.__setattr__(
+            self,
+            "flagtree_submodule_dir",
+            os.path.join(self.flagtree_root_dir, "third_party"),
+        )
+        object.__setattr__(self, "activated_module", self._activate_device_module(self.flagtree_backend))
```
sgjzfzzf: Why cannot we set attribute directly here?

python/setup_tools/utils/tools.py
```diff
@@ -10,12 +10,58 @@
...
+
+    def _activate_device_module(self, backend, suffix=".py"):
+        backend = "default" if not backend else backend
```
sgjzfzzf:
```suggestion
        backend = backend or "default"
```

python/setup_tools/utils/tools.py
```diff
@@ -10,12 +10,58 @@
...
+        module_path = Path(os.path.dirname(__file__)) / backend
+        module_path = str(module_path) + suffix
+        spec = importlib.util.spec_from_file_location("module", module_path)
+        module = importlib.util.module_from_spec(spec)
+        try:
+            spec.loader.exec_module(module)
+        except Exception:
```
sgjzfzzf: Don't "except Exception" here directly. That would make us lose other unexpected errors. Check the possible exception types here and assign specific ones.

Galaxy1458 (approved):
LGTM

PR #345, opened 2026-02-04
[TLE_Raw] Move some ops out of DSLregion

SabYic:
ops includes tle::ExtractSizesOp, tle::ExtractStridesOp,tle::ExtractOffsetOp, tle::ExtractAllocatedPtrOp,ExtractAlignedPtrOp


SabYic:
ready for review. I’d appreciate any suggestions. The code seems somewhat messy.

sgjzfzzf (changes requested):
Don't implement `ConvertArgToMemDesc` like this.

You could still try to match the `DSLRegionOp`, and check its inputs and outputs to modify them. The memory allocations and deallocations are paired by `DSLRegionOp`, so it would be better to match it instead of `ExtractOp`s and `PackOp`s.

`findNearestNextDSLRegionOp` doesn't look robust enough. You shouldn't assume that caller and callee relationship by their positions. Follow def-usage chain instead of that.

third_party/tle/triton_tle_raw.cc
```diff
@@ -25,6 +25,27 @@ SmallVector<Value> flatten(TritonOpBuilder &builder,
...
+  for (Type singletype : unconvertTypes) {
+    if (auto ptrType = dyn_cast<RankedTensorType>(singletype)) {
+      size_t rank = ptrType.getRank();
+      SmallVector<Type> fieldTys = tgts.take_front(3 + rank * 2);
+      resultTypes.push_back(LLVM::LLVMStructType::getLiteral(
+          builder.getContext(), fieldTys, /*packed=*/false));
+      tgts = tgts.drop_front(3 + rank * 2);
+    } else {
+      // struct type
+      resultTypes.push_back(tgts.front());
```
sgjzfzzf:
```suggestion
      resultTypes.push_back(std::move(tgts.front()));
```

third_party/tle/triton_tle_raw.cc
```diff
@@ -80,15 +101,49 @@ tle::DSLRegionOp createTLERawRegionByLLVMFunc(
...
+      converted_outputs.append(std::move(rets));
+    }
+  }
   SmallVector<Type> outputTys = llvm::map_to_vector(
       outputs, [](Value value) -> Type { return value.getType(); });
-  SmallVector<Value> operands = llvm::to_vector(
-      llvm::concat<Value>(SmallVector<Value>(outputs.begin(), outputs.end()),
-                          SmallVector<Value>(inputs.begin(), inputs.end())));
 
+  // SmallVector<Value> operands = llvm::to_vector
```
sgjzfzzf: Unused line.

third_party/tle/triton_tle_raw.cc
```diff
@@ -80,15 +101,49 @@ tle::DSLRegionOp createTLERawRegionByLLVMFunc(
...
+    SmallVector<Type> funcArgTypes =
+        llvm::map_to_vector(func.getArguments(), [](BlockArgument arg) -> Type {
+          return arg.getType();
+        });
+    TypeRange tgts = funcArgTypes;
+    for (Value src : outputs) {
+      SmallVector<Value> rets =
+          tle::protocol::SignaturePattern::apply(self, tgts, src);
+      converted_outputs.append(std::move(rets));
+    }
```
sgjzfzzf: Use map and concat to aggregate this for loop.

third_party/tle/triton_tle_raw.cc
```diff
@@ -80,15 +101,49 @@ tle::DSLRegionOp createTLERawRegionByLLVMFunc(
...
+    SmallVector<Type> funcArgTypes =
+        llvm::map_to_vector(func.getArguments(), [](BlockArgument arg) -> Type {
+          return arg.getType();
+        });
+    TypeRange tgts = funcArgTypes;
+    for (Value src : inputs) {
+      SmallVector<Value> rets =
+          tle::protocol::SignaturePattern::apply(self, tgts, src);
+      converted_inputs.append(std::move(rets));
+    }
```
sgjzfzzf, line 126: The same as above.

third_party/tle/dialect/lib/Transforms/ConvertArgToMemDesc.cpp
```diff
@@ -59,95 +71,75 @@ TleArgConversion::TleArgConversion(MLIRContext *context)
...
-    }
+  for (Operation *toReplace : targets) {
+    rewriteExtractWithMappedInput<tle::ExtractSizesOp>(toReplace, mapper,
+                                                       rewriter);
+    rewriteExtractWithMappedInput<tle::ExtractStridesOp>(toReplace, mapper,
+                                                         rewriter);
+    rewriteExtractWithMappedInput<tle::ExtractOffsetOp>(toReplace, mapper,
+                                                        rewriter);
+    rewriteExtractWithMappedInput<tle::ExtractAlignedPtrOp>(toReplace, mapper,
+                                                            rewriter);
```
sgjzfzzf: It would be helpful if we could merge them into one statement like
```cpp
rewriteExtractWithMappedInput<tle::ExtractSizesOp, tle::ExtractStridesOp, tle::ExtractOffsetOp,tle::ExtractAlignedPtrOp>(toReplace, mapper, rewriter);
```

third_party/tle/triton_tle_raw.cc
```diff
@@ -139,16 +190,22 @@ tle::DSLRegionOp createTLERawRegionByLLVMFunc(
...
+  builder.setInsertionPointAfter(dslRegionOp);
+  SmallVector<Value> finalResults;
+  if (dslRegionOp.getNumResults() > 0) {
+    TypeRange tgts = outputTys;
+    for (Value result : dslRegionOp.getResults()) {
+      SmallVector<Value> rets =
+          tle::protocol::ReturnPattern::apply(self, tgts, result);
+      finalResults.append(std::move(rets));
+    }
+  }
```
sgjzfzzf: If statement is unnecessary, and try to aggregate it.

third_party/tle/dialect/lib/Transforms/ConvertArgToMemDesc.cpp
```diff
@@ -59,95 +71,75 @@ TleArgConversion::TleArgConversion(MLIRContext *context)
...
+        rewriter.create<ttg::LocalStoreOp>(ex.getLoc(), ex.getInput(), allocOp);
+        auto newAligned = rewriter.create<tle::ExtractAllocatedPtrOp>(
+            ex.getLoc(), ex.getResult().getType(), allocOp);
+        mapper.map(ex.getInput(), allocOp.getResult());
+        rewriter.replaceOp(ex, newAligned.getResult());
+        rewriter.setInsertionPointAfter(op);
+        toDeallocOps.push_back(allocOp);
+        hasConversion = true;
+      }
+    }
```
sgjzfzzf: Don't depend on specific op types. You should use a set to record all owners of the `dslRegionOp`'s inputs, and then convert them one by one. You shouldn't assume every input will generate a `ExtractAllocatedPtrOp`. That's dangerous and not robust if we need to extend our current system.

third_party/tle/dialect/lib/Transforms/ConvertArgToMemDesc.cpp
```diff
@@ -59,95 +71,75 @@ TleArgConversion::TleArgConversion(MLIRContext *context)
...
+      targets.push_back(defOp);
+    }
+  }
+  SmallVector<tle::PackOp> packs;
+  for (Value res : op->getResults()) {
+    for (OpOperand &use : res.getUses()) {
+      if (auto pack = dyn_cast<tle::PackOp>(use.getOwner()))
+        packs.push_back(pack);
     }
   }
```
sgjzfzzf: If the only usage of `packs` is the following loop, why u cannot aggregate these loops?

third_party/tle/utils/lib/Protocol.cpp
```diff
@@ -77,17 +77,29 @@ SmallVector<Value> LLVMStructurePattern::apply(TritonOpBuilder &builder,
...
+       }));
+
+  bool is_flattened =
+      (size == 2 * rank + 3 &&
+       llvm::all_of(types.take_front(2),
+                    [](const Type &ty) -> bool {
+                      return isa<LLVM::LLVMPointerType>(ty);
+                    }) &&
+       llvm::all_of(types.drop_front(2),
+                    [](const Type &ty) -> bool { return ty.isInteger(64); }));
```
sgjzfzzf: When do we need to check this? I remember the standard layout would be like
```
llvm.struct<llvm.ptr, llvm.ptr, i32, array<?xi32>, array<?xi32>
```

third_party/tle/dialect/lib/Transforms/ConvertArgToMemDesc.cpp
SabYic: only assume the existence of ExtractAllocatedPtrOp when dealing with a TensorType. In addition, I need to identify an operation that can serve as an anchor to determine where to insert ttg::alloc, and I haven’t found a better approach so far.

sgjzfzzf: Before the first usage of op or after its definition.

third_party/tle/dialect/lib/Transforms/ConvertArgToMemDesc.cpp
```diff
@@ -25,7 +25,24 @@ namespace ttg = mlir::triton::gpu;
...
+    rewriter.replaceOp(ex, newEx->getResults());
+  }
+}
 
+template <typename... OpTys>
+static void rewriteExtractWithMappedInput(Operation *toReplace,
+                                          IRMapping &mapper,
+                                          PatternRewriter &rewriter) {
+  (rewriteOne<OpTys>(toReplace, mapper, rewriter), ...);
+}
```
sgjzfzzf, line 65: Let `rewriteOne` return a bool value, so in `rewriteExtractWithMappedInput` you could use short-circuit evalution to skip unnecessary branches.

third_party/tle/triton_tle_raw.cc
```diff
@@ -80,15 +108,44 @@ tle::DSLRegionOp createTLERawRegionByLLVMFunc(
...
+  {
+    OpBuilder::InsertionGuard guard(builder);
+    TypeRange tgts = funcArgTypes;
+    SmallVector<Value> rets;
+    llvm::for_each(inputs, [&](mlir::Value v) {
+      auto xs = tle::protocol::SignaturePattern::apply(self, tgts, v);
+      rets.append(xs);
+    });
+    converted_inputs.append(std::move(rets));
+  }
```
sgjzfzzf: emm, I checked LLVM and realized that it doesn't support map-fold programming style

Go back to the for-loop may look better
Sorry for that😭

sgjzfzzf (approved):

sunnycase (approved):
LGTM

PR #347, opened 2026-02-04
[CI] Change runner label from iluvatar to flagtree-iluvatar

zhzhcookie:
Because flagos-ai added some runner on label iluvatar, so FlagTree changes the runner label to flagtree-iluvatar.


Galaxy1458 (approved):
LGTM

PR #357, opened 2026-02-12
[TLERaw] replace DslRegionOp with LLVM::CallOp

sgjzfzzf:
This PR is used to remove `DSLRegionOp`, which would prepare for the support for native language integration. `LLVM::CallOp` is used to call the TLERaw kernel instead of inlining the whole kernel inside the body. We use `always_inline` to defer the inline to the LLVM phase.


zhzhcookie (approved):
LG

sunnycase (approved):
LGTM

PR #387, opened 2026-03-05
[TLERaw] Revert `LLVM::CallOp`-Related Modification

sgjzfzzf:
The current implementation has known bugs on `topk` implementation. We revert our code back to the previous implementation and modified CIs so it could intercept the error correctly.


sunnycase (approved):
LGTM

zhzhcookie (approved):
LG

PR #462, opened 2026-03-20
[TLE] Add prefill and decode triton kernels for causal conv1d

small-cat:
This pull request introduces prefill and decode triton kernels for causal conv1d implemented with flagtree triton tle in `python/tutorials/tle`.


zhzhcookie (approved):
LG

PR #474, opened 2026-03-25
[BACKEND] Fix issue of editable mode for sunrise backend

SabYic:


sgjzfzzf (dismissed):

setup.py
zhzhcookie, line 726: How should these commented-out code sections be handled?

PR #538, opened 2026-04-21
[DOC] Fix install-from-source instructions for Triton 3.4+ branches

Schopenhauer-loves-Hegel:
## Summary

The install-from-source section in README.md only mentions `cd python` before running `pip install`. This works for Triton 3.1/3.2/3.3 branches where `setup.py` is under `python/`, but fails on Triton 3.4+ branches (triton_v3.4.x, triton_v3.5.x, triton_v3.6.x) where `setup.py` has been moved to the project root.

Running `cd python && pip install .` on these branches gives:
```
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
```

## Changes

Clarified the install instructions to distinguish between:
- Triton 3.1/3.2/3.3: install from `python/` directory
- Triton 3.4+: install from project root directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,15 @@ General building and installation procedure (Recommended for environments with g
 ```shell
 # Set FLAGTREE_BACKEND using the backend name from the table above
 export FLAGTREE_BACKEND=${backend_name}  # Do not set it on nvidia/amd/triton-shared
-cd python  # Need to enter the python directory for Triton 3.1/3.2/3.3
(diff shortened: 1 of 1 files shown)


zhzhcookie:
Thanks for the suggestion. We will update readme to avoid ambiguity in the build command.
Refer to https://github.com/flagos-ai/FlagTree/pull/539

PR #582, opened 2026-05-12
[Ascend] Align Ascend do_bench_npu with CUDA's do_bench

wangziy1:
[FEAT] Align do_bench_npu with CUDA's do_bench
1. add a new ascendC kernel to clear L2 and avoid time calculation error because of torch API
2. add return mode to get "min/max/median/mean" results
3. add matmul kernel before benchmark to stabilize Ascend NPU frequency
4. change do_bench_npu warmup/active from iteration count to duration (ms)

### DIFF
--- a/third_party/ascend/ClearL2Cache/CMakeLists.txt
+++ b/third_party/ascend/ClearL2Cache/CMakeLists.txt
@@ -1,0 +1,44 @@
+# Copyright 2026- Xcoresigma Technology Co., Ltd
+cmake_minimum_required(VERSION 3.16)
+find_package(ASC REQUIRED)
+project(ClearL2Cache LANGUAGES CXX ASC)
+find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
+
+# pybind11
+execute_process(
+    COMMAND ${Python3_EXECUTABLE} -c "import pybind11; print(pybind11.get_cmake_dir())"
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    OUTPUT_VARIABLE PYBIND11_CMAKE_PREFIX_PATH
+    ERROR_QUIET
+)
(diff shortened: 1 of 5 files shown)


third_party/ascend/ClearL2Cache/CMakeLists.txt
```diff
@@ -0,0 +1,123 @@
...
+    "-Wl,-rpath,${TORCH_NPU_LIBRARIES}"
+    "-Wl,-rpath,${CANN_LIB_DIR}"
+    "-Wl,-z,now"
+)
+
+# POST_BUILD  
+execute_process(
+    COMMAND ${Python3_EXECUTABLE} -c "import site; print(site.getsitepackages()[0])"
+    OUTPUT_VARIABLE PYTHON_SITE_PACKAGES OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET)
+find_program(PATCHELF_EXECUTABLE patchelf REQUIRED)
```
neoskypig: Why did you add this patchelf?

wangziy1: The root cause has been identified as a dynamic linking resolution failure triggered by the module import chain. Upon import triton, the execution of __init__.py cascades through triton/backends/ascend/__init__.py and testing.py, which unconditionally imports clear_l2 at the module level. During the dynamic loading of clear_l2.so, the system linker (ld.so) fails to resolve the transitive dependency libtorch_python.so due to missing runtime search paths (DT_RUNPATH/RPATH) in the ELF binary, ultimately resulting in an **ImportError**, Like this: 
img_v3_0211m_0efd2a29-cee6-4ee0-9364-282513cd355g

wangziy1: After decoupling torch_npu, patchelf is **no longer needed**.

zhzhcookie (approved):
LG

PR #588, opened 2026-05-14
[TLE] Avoid redundant downloads in trtllm topk loader

kuoihao:
The original code downloaded the same source file multiple times when
different prefill_threads values (512 and 1024) were used, because
lru_cache treats different arguments as separate cache keys.

This caused CI failures when the second download timed out due to
unstable network conditions.

Fix by extracting download logic into a separate cached function
_download_trtllm_indexer_topk_source() with no parameters, ensuring
the source is downloaded only once regardless of prefill_threads.


sunnycase (approved):
LGTM

zhzhcookie (approved):
LG

PR #629, opened 2026-05-28
[TLE] Support mixed pipe effects and encoding rematerialization

sunnycase:
## Summary

Compared with `triton_3.6.x`, this PR adds the TLE compiler support needed to express FlashQLA-style warp-specialized pipelines with one logical dataflow model instead of transport-specific synchronization fragments.

### Mixed Pipe

A mixed pipe is a normal `tle.pipe` whose slot contains multiple payload fields produced by different transports. For example, one field can be filled by TMA, another by cp.async-style copy, and another by CUDA-core shared-memory stores through `tle.gpu.local_ptr` and `tl.store`. Users still create one logical pipe, produce all fields between `writer.acquire(iter)` and `writer.commit(iter)`, and let the compiler infer each field transport from producer-side IR.

The motivation is to split pipes by logical lifecycle and reader protocol, not by transport. Without mixed pipe support, a kernel has to decompose one logical chunk into multiple pipes such as TMA pipes, scale/meta local-store pipes, and cp.async pipes. That duplicates ready/free barriers, increases shared-memory synchronization overhead, and obscures the intended producer/consumer graph. Mixed pipe keeps one slot, one phase, one reader set, and one ready/free protocol while preserving field-level provenance and compile-time hazard checks.

To use it, allocate one shared-memory field per payload component with the same leading `capacity` dimension, pass all fields to one `tle.pipe(...)`, write each field with its natural operation, and subscribe readers with `pipe.reader(name, fields=(...))` when they only need a subset. Transport is not encoded in pipe names, field names, or user attributes; unsafe opaque shared-memory effects or unprovable source-order hazards fail at compile time.

### Compiler Support

- Adds TLE pipe effect and memory provenance analysis so pipe transports are inferred from IR, including mixed TMA/local-store/cp.async payloads, participant commits, and early-release semantics with fail-fast diagnostics for unsafe interleaves.
- Moves pipe-to-NVWS commit transport selection to original-IR payload-window analysis, so one-shot multi-field TMA pipe setup is independent of rewrite order and preserves source-order semantics.
- Adds TLE-owned encoding rematerialization for store layout demands, enabling MMA/local-pointer store paths to rematerialize address and mask computations in the required encodings without op-specific lowering workarounds.
- Adds TLE-only WGMMA stage predication in TritonGPU pipelining, preserving accumulator semantics for dynamically predicated warp-group dot stages.
- Updates `tle.md` and `tle_cn.md` with mixed pipe motivation, semantics, usage rules, and current NVIDIA lowering constraints.
- Adds TLE-owned regression coverage for mixed pipe lowering, NVWS conversion, encoding rematerialization, one-shot multi-field TMA pipes, WGMMA stage predicates, and related fail-fast diagnostics.

### Dynamic WGMMA Stage Predication

The main correctness bug fixed here is in TritonGPU pipelining for dynamically predicated WGMMA stages. When software pipelining creates a non-constant stage predicate around `ttng.warp_group_dot`, treating the dot like a regular predicated operation is not sufficient: a false stage predicate must skip the WGMMA issue entirely and preserve the incoming accumulator. Otherwise a boundary/prologue/epilogue stage can still issue WGMMA outside the valid stage and either update the accumulator incorrectly or fail during later lowering/runtime execution.

This PR fixes that path by lowering dynamically predicated `ttng.warp_group_dot` under a TLE guard into an explicit `scf.if`: the true branch issues the WGMMA and yields its result, while the false branch yields the original accumulator. This preserves the SSA accumulator contract and makes the false path a real control-flow skip, not a masked WGMMA issue.

Impact scope:

- Affected: NVIDIA/TLE kernels that use TritonGPU pipelining with WGMMA and dynamic stage predicates, especially warp-specialized kernels whose loop prologue/epilogue predicates are not compile-time constants.
- Observed reproducer: `flash-linear-attention` gated delta product model test, which exercises a pipelined WGMMA path through the model forward/backward stack.
- Not affected: non-WGMMA dot paths, constant-true pipeline stages, and non-TLE builds; the native Triton hook is guarded by `__TLE__`.
- Regression coverage: `third_party/tle/test/GPU/test_tle_wgmma_stage_predication.mlir` checks that dynamic WGMMA predicates become explicit `scf.if` control flow and that the false branch yields the accumulator.

## Why It Matters

This makes TLE pipe lowering deterministic and analyzable for kernels that mix TMA, local stores, cp.async, WGMMA, and MMA-encoded stores inside the same warp-specialized pipeline. The compiler now derives transport contracts from IR ownership and source order, keeps TLE behavior isolated from upstream-owned paths, and rejects unsafe cases at compile time instead of relying on lowering-order assumptions.

## Validation

- `conda run -n flagtree ninja -C build/cmake.linux-x86_64-cpython-3.10`
- `conda run -n flagtree ninja -C build/cmake.linux-x86_64-cpython-3.10 triton-opt triton check-triton-tle-lit-tests`
- `conda run -n flagtree pytest python/test/tle/unit/test_tle.py -q`
- `PYTHONPATH=/root/repos/flagtree/python:/root/repos/FlashQLA TRITON_CACHE_DIR=/tmp/flagtree_tle_guarded_cache conda run -n flagtree python build/flashqla_gdr_tle_pipe.py --correctness --bench --warmup 50 --rep 200`
- `PYTHONPATH=/root/repos/flagtree/python:/root/repos/flash-linear-attention TRITON_CACHE_DIR=/tmp/flagtree_fla_gdp_cache conda run -n flagtree pytest tests/models/test_modeling_gated_deltaproduct.py -q -s --tb=short`
- Sparse MLA TLE pipe bf16 forward/decode correctness matrix
- Sparse MLA TLE-Pipe prefill performance check on H800: current head is within ~0.2%-4.0% of the 2026-05-19 baseline across SKV 8192-131072, with large-SKV cases around ~2% slower.
- Direct TLE pipe fp16 forward smoke test


zhzhcookie (approved):
LGTM

PR #634, opened 2026-05-29
[Ascend] Fix sync issues in do_bench_npu

wangziy1:
[TLE] Fix sync issues in do_bench_npu.
1. Fix synchronisation order when calculating warmup count. The original code called fn() in the estimation loop without synchronisation, which caused event recording to be inaccurate. Added torch.npu.synchronize() after each call to ensure all kernels complete before recording events.
2. Move matmul frequency-boosting outside the profiler scope. The MatMul used to raise NPU frequency should run on NPU (not CPU), and must be placed before the profiler context to avoid polluting profile results with irrelevant kernel data.


neoskypig:
@wangziy1 please add the detailed commit message to explain the reason

zhzhcookie (approved):
LGTM

PR #648, opened 2026-06-04
[KMCompiler][JIT] fix performance regression caused by explicit annotate constexpr value as int

zhongsanming:
fix a performance issue introduced upstream triton, where an explicit type annotation of `: int` will overwrite constexpr when generating code, which further cause the coalesce pass failed to generate optimized IR. 

This affect all triton version starting from 3.4, and most of the op in FlagGems repo.

See also: 

[ triton-lang/triton#9834](https://github.com/triton-lang/triton/issues/9834#issuecomment-4136722453)

 triton-lang/triton#6152


zhzhcookie (approved):
LGTM

PR #649, opened 2026-06-04
[TLE] Feature/extrac tile strides new

lzllx123:
### Summary:

1. [TLE]Add optional strides parameter to tle.extract_tile and tle.insert_tile, enabling strided tile extraction/insertion.
2. Adapted for NVIDIA and GCU backends.

### API Changes:

1. Python: extract_tile(x, index, tile_shape, strides=None) / insert_tile(x, tile, index, strides=None)
2. MLIR: Tle_ExtractTileOp and Tle_InsertTileOp added OptionalAttr<DenseI64ArrayAttr>:$strides
3. Backward compatible: defaults to tile_shape when omitted

### Performance:
New tutorials optimized with tle.extract_tile / tle.insert_tile:

1. 05-glu.py (GLU)
2. 06-2D_Depthwise_Conv.py (2D Depthwise Conv)
3. 07-causal-conv1d.py (Causal Conv1D)
4. 08-rope.py (RoPE)

### CI: 
Added lightweight correctness tests for tutorials 05-08 in hopper-build-and-test.yml.

PR #655, opened 2026-06-05
[debug] Instrument TLE cluster GEMM CI failure

sunnycase:
Temporary CI debug PR for the `04-cluster-gemm.py` Hopper failure in PR #649.

Changes:
- Print runtime/cache/autotune-candidate diagnostics when `FLAGTREE_CI_DEBUG_CLUSTER_GEMM=1`.
- Run the suspected `BK=256` remote config under `CUDA_LAUNCH_BLOCKING=1` before the original full default run in the triton_v3.6.x Hopper workflow.

This PR is intended for diagnostics only and should be closed after collecting logs.


sunnycase:
CI debug summary for PR #649:

- Original PR #649 Hopper rerun still fails in `python/tutorials/tle/04-cluster-gemm.py` with `CUDA error: misaligned address` at `_autotune_config -> torch.cuda.synchronize()` immediately after baseline `triton_gemm` autotune.
- Debug run with `CUDA_LAUNCH_BLOCKING=1` passed. Candidate 1 for remote autotune was `KernelConfig(bm=32, bn=512, bk=256, num_warps=4, num_stages=2)` and completed launch/sync/bench successfully.
- Debug run without `CUDA_LAUNCH_BLOCKING`, but with per-candidate `print(..., flush=True)`, also passed; all 16 remote candidates completed.

Current inference: the failure is timing-sensitive / host-side perturbation-sensitive, not a deterministic bad config. The first remote candidate remains the likely failing launch in the uninstrumented script, but adding prints is enough to hide it on the Hopper runner.

PR #665, opened 2026-06-08
[Hints] Optimize hint manager lookup

sunnycase:
## Summary

- Attach parsed line hint metadata to `CodeGenerator` for kernels and nested JIT functions so hint lookups do not reparse the JIT source for every node.
- Cache NVIDIA hint comment tokenization by source text while returning an independent dictionary to callers.
- Add focused unit coverage for non-reparsing hint lookup and cached source mapping isolation.

## Release Compile-Time Measurement

Build/config:

- Built and installed with `CMAKE_BUILD_TYPE=Release` in `build/cmake.release.linux-x86_64-cpython-3.10`.
- Verified `build/cmake.release.linux-x86_64-cpython-3.10/CMakeCache.txt` contains `CMAKE_BUILD_TYPE:STRING=Release`.
- Target: NVIDIA H800, `GPUTarget("cuda", 90, 32)`.

Benchmark:

- Synthetic Triton kernel with 96 `tl.load(...)  # @hint: shared_memory` call sites.
- Measured compile only via `triton.compile(ASTSource(...), target=GPUTarget("cuda", 90, 32))`; the kernel is not launched.
- Used `TRITON_ALWAYS_COMPILE=1` plus isolated `TRITON_CACHE_DIR` values for before/after runs.
- 1 warmup compile + 7 measured compiles, same release `libtriton.so` for both revisions.

Results:

| Revision | Median compile time | Mean compile time |
| --- | ---: | ---: |
| Before (`259ae42a0`) | 1.550 s | 1.572 s |
| After (`dc5b2f88b`) | 0.819 s | 0.778 s |

This is a 1.89x median speedup, or 47.2% lower median compile time, on this hint-heavy release compile benchmark.

## Testing

- `conda run -n flagtree pytest python/test/unit/language/test_hint_manager.py -q`


zhzhcookie (approved):
LGTM

i3wanna2 (approved):
LGTM

PR #669, opened 2026-06-08
[Ascend] feat: update do_bench_npu

winter77-x:

### DIFF
--- a/third_party/ascend/backend/testing.py
+++ b/third_party/ascend/backend/testing.py
@@ -28,15 +28,17 @@ import triton.runtime as runtime
 from triton._C.clear_l2 import do_bench_clear
 
 
-def do_bench_npu(funcs, warmup=25, active=100, prof_dir=None, clear_l2_cache=False, keep_res=False, collect_prof=True,
-                 return_mode="mean", quantiles=None):
+def do_bench_npu(funcs, warmup=5, active=30, prof_dir=None, clear_l2_cache=False, keep_res=False):
     import torch
     import torch_npu
 
-    assert return_mode in ["min", "max", "mean", "median", "all"]
-
     if not isinstance(funcs, list):
         funcs = [funcs]
+
+    # warmup kernel
+    for fn in funcs:
+        fn()
+        torch.npu.synchronize()
 
     experimental_config = torch_npu.profiler._ExperimentalConfig(
         aic_metrics=torch_npu.profiler.AiCMetrics.PipeUtilization,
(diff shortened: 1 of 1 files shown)


third_party/ascend/backend/testing.py
```diff
@@ -52,74 +54,42 @@ def do_bench_npu(funcs, warmup=25, active=100, prof_dir=None, clear_l2_cache=Fal
...
+        )
         torch_path = os.path.join(base_path, f"prof_{timestamp}_{process_name}-{pid}")
 
     if clear_l2_cache:
-        device = triton.runtime.driver.active.get_current_device()
-        stream = triton.runtime.driver.active.get_current_stream(device)
         buffer = runtime.driver.active.get_empty_cache_for_benchmark()
-        do_bench_clear(buffer.data_ptr(), buffer.numel(), stream)
+        buffer = buffer.float()  # to avoid type cast
+        buffer.sum()
```
neoskypig, line 65: usually，we use a CPU "wrapper" to call a triton-kernel，it is probably that  using the "torch.sum"  in the wrapper function， and it will cause the time is wrong because this "torch.sum" time is captured by msprof。

third_party/ascend/backend/testing.py
```diff
@@ -52,74 +54,42 @@ def do_bench_npu(funcs, warmup=25, active=100, prof_dir=None, clear_l2_cache=Fal
...
-    n_repeat = min(30, max(1, int(active / estimate_ms)))
-
-    # cal warmup num
-    total = n_warmup + n_repeat
-    print(f"total={total}, n_warmup={n_warmup}, n_repeat={n_repeat}")
-
-    # Run for 300 μs to raise the frequency to 800.
-    mat_a = torch.randn(4096, 4096).to(dtype=torch.bfloat16).npu()
-    mat_b = torch.randn(4096, 4096).to(dtype=torch.bfloat16).npu()
-    mat_c = torch.matmul(mat_a, mat_b)
```
neoskypig, line 93: according to experience, it needs a ">= 300us" kernel which make NPU works on a high frequency。
this method is used by some open-source libraries, e.g. bytedance's "mojoOpset".

third_party/ascend/backend/testing.py
winter77-x: In the do_bench_npu function, buffer.sum() is used to forcefully clear the NPU’s L2 cache before each iteration of the benchmark (when clear_l2_cache=True).

third_party/ascend/backend/testing.py
winter77-x: The warmup in do_bench_npu is implemented in two stages:
1. Pre‑warmup (1 call per function) – ensures the runtime environment is “ready” without being measured
2. Profiled warmup (warmup calls per function) – allows the device to reach a steady state, but these calls are excluded from the final results.

PR #683, opened 2026-06-10
[TLE][MTHREADS] Support TLE Lite primitives 

Kylin1207:
## MTHREADS backend support for the main TLE Lite primitives in this patch:

  - `tle.extract_tile(x, index, tile_shape)`
    - Extracts a tile of shape `tile_shape` from a register tensor at a tile-grid index.
    - Supported index forms:
      - Multi-dimensional static (tuple/list of int/constexpr), linearized at compile time.
      - Scalar static (int/constexpr), treated as the linearized tile index.
      - Scalar dynamic (scalar i32/i64 `tl.tensor`), lowered through the shared-memory relay path.
      - Multi-dimensional dynamic (tuple/list containing `tl.tensor`), row-major linearized automatically in the frontend.
    - `tile_shape` must be compile-time constants, match the source rank, and divide the source shape evenly along each dimension.

  - `tle.insert_tile(x, tile, index)`
    - Inserts a register tile into a source tensor at a tile-grid index; result type equals the source type.
    - Index forms and constraints are the same as `extract_tile`; the tile shape is derived from the tile operand and must divide the source shape evenly.

  - `tle.cumsum(input, axis=0, reverse=False, dtype=None)`
    - Returns `(exclusive_sum, total_sum)` over a rank-1 tensor in one call:
      `exclusive_sum[i] = sum(input[:i])`, `total_sum = sum(input)`.
    - `reverse=True` produces the reverse-exclusive prefix with the same total.
    - dtype promotion follows the shared frontend: int8/int16 -> int32, uint8/uint16 -> uint32, bf16 -> fp32; int32/fp16/fp32 stay unchanged.
    - Limitations: only rank-1 tensors with axis=0 and static positive extent are supported; scalar input short-circuits to `(0, input)` in the frontend.

## Lowering path

  Key differences from native Triton:
  - All three primitives are backend-local `musa_tle` ops (`musa_tle.extract_tile`, `musa_tle.insert_tile`, `musa_tle.exclusive_cumsum`); the shared `tle` dialect is not linked into the mthreads build.
  - Tile ops keep their op identity through TTIR -> TTGIR and lower in MUSATLEToLLVM either through register shuffles (aligned static index) or a shared-memory relay (dynamic index), with scratch sized in `AllocateSharedMemory`.
  - `musa_tle.exclusive_cumsum` is lowered at TTGIR level by `tritonmusa-tle-lower-exclusive-cumsum` into one add-combiner `tt.scan`, an elementwise subtract for the exclusive result, and an index-select `tt.reduce` for the total.

## Performance Data
  Benchmark source:
  -  third_party/mthreads/python/test/unit/tle/benchmark/test_per_token_group_quant_fp8.py

  Environment:
  - Driver: 4.3.5
  - SDK: 5.1.0
  - Torch: 2.9.0

  TLE vs Triton Native: 1.25x SpeedUp on large shape
  test cmd: `python  third_party/mthreads/python/test/unit/tle/benchmark/test_per_token_group_quant_fp8.py --dtype bf16 --shape 1024,11008`

### DIFF
--- a/.github/workflows/mthreads-build-and-test.yml
+++ b/.github/workflows/mthreads-build-and-test.yml
@@ -46,3 +46,6 @@ jobs:
         run: |
           set -x
           python3 -m pytest -s third_party/mthreads/python/test/unit --device musa
+          # TLE tutorials unit tests
+          python3 python/tutorials/tle/01-fft.py --only_unit_test
+          python3 python/tutorials/tle/03-topk.py --only_unit_test
--- a/python/tutorials/tle/01-fft.py
+++ b/python/tutorials/tle/01-fft.py
@@ -292,7 +292,7 @@ if _HAVE_CUTILE:
 def cutile_fft(x: torch.Tensor) -> torch.Tensor:
     if not _HAVE_CUTILE:
         raise RuntimeError("cuda.tile/cupy not available")
-    assert x.is_cuda, "input must be on CUDA"
+    assert x.device.type == DEVICE.type, "input must be on device"
     assert x.ndim == 2, "input must be 2D (M, N)"
     m, n = x.shape
     if not _is_power_of_two(n):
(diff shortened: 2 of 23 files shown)


zhzhcookie (approved):
LGTM

PR #693, opened 2026-06-11
[RPU][CI] Stop a wedged launch_kernel dispatch from failing the board smoke

xup16:
## Problem
When a `launch_kernel` dispatch on the RPU board hangs, the board smoke's 120s subprocess timeout only SIGKILLs the direct child (`sudo`). sudo runs the real `launch_kernel_runner` in its own pty session, so it survives as an orphan that keeps `/dev/rpu` open and leaves `run_work` stuck in the RPU driver queue. Every later case then opens the device, hits `run_work was already in queue`, and times out too -- one hung dispatch turns into `0/N passed`, and only a board reboot clears it. (Observed in CI: all 11 board-smoke cases `TimeoutExpired` while the unit suite stayed green.)

## Fix
- **`lk_board_smoke.py`**: run each dispatch via `Popen(start_new_session=True)`; on timeout, kill the whole process group and surgically reap any runner still matching the case's unique work dir, so a concurrent job on a shared board is never touched. The happy path still returns a `CompletedProcess` unchanged.
- **`rpu3.6` workflow**: before testing, reap any `launch_kernel_runner` alive far longer than the 120s per-case cap (a healthy dispatch finishes in seconds), clearing a pre-existing wedge without killing a concurrent dispatch.

## Verification
- `py_compile` + workflow YAML parse OK.
- Unit-tested the happy path (returns `CompletedProcess`) and the timeout teardown (process group fully reaped, no surviving runner).
- Board smoke is green 11/11 on the physical board; this patch leaves the dispatch happy-path identical and only changes timeout teardown.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/rpu3.6-build-and-test.yml
+++ b/.github/workflows/rpu3.6-build-and-test.yml
@@ -78,5 +78,15 @@ jobs:
           ssh -tt -o BatchMode=yes -o StrictHostKeyChecking=no rpuboard 'set -e
             cd ~/FlagTree-ci
             source ~/env.sh
+            # Reap any leaked launch_kernel_runner from an earlier wedged run:
+            # an orphan holding /dev/rpu leaves run_work stuck in the RPU queue
+            # and makes every dispatch here time out. Only kill ones alive far
+            # longer than the 120s per-case cap -- a healthy dispatch finishes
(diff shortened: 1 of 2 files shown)


zhzhcookie (approved):
LGTM

PR #700, opened 2026-06-15
[TLE-Raw] Add DSLRegion metadata for deferred vendor lowering and fix redundant-copy removal

i3wanna2:
## Summary

 This PR extends `tle.dsl_region` with metadata (`region_dialect`, `arg_dialect`, `output_operand_indices`, optional `hint`) to support vendor-specific DSL regions. It also implements **deferred lowering** for the NVIDIA path as a primary reference implementation, while keeping the existing **eager** CUDA and MLIR paths working.

In deferred mode, external DSL code (for example a CUDA `.cu` file) is not compiled during kernel tracing. Instead, the compiler records the source and emits a stub `dsl_region`. The real region body is materialized later in `make_llir`, before `dsl_region` is inlined into the parent kernel.

 ## Changes

  • Refactor Python/C++ creation path (call / call_smem, CUDA runtime, source_store, deferred create API)
  • Keep ConvertArgToMemDesc as a pure conversion pass
  • Fix RemoveRedundantCopy to use output_operand_indices for shared-memory redundant-copy elimination
  • Auto-derive alias indices from LLVM return analysis when output_indices is not provided
  • Add minimal TOPS Python adapter for compatibility with the shared creation API

  ---
  
  ## Eager vs Deferred

### Eager mode (default)

When the kernel is traced and `tle_raw.call()` or `tle_raw.call_smem()` is reached, the external source is compiled immediately (clang for CUDA, the MLIR pipeline for MLIR Python EDSL). The compiler then builds a full `dsl_region` body with a real LLVM call to the external device function.

```
Trace  →  compile external source  →  full dsl_region with LLVM call
```

`output_operand_indices` can be inferred automatically from LLVM return-alias analysis when the user does not pass `output_indices`.

### Deferred mode (`deferred=True`)

During tracing, the compiler only registers the external source in a pending store and emits a **stub** `dsl_region`. The stub body simply yields the aliased input operands unchanged. A `tle_raw.source_id` attribute links the op to the pending source.

The external source is compiled once in `make_llir`. The `MaterializeDeferredRaw` pass replaces the stub body with the real LLVM call, then removes `tle_raw.source_id`.

```
Trace      →  register source  →  stub dsl_region (yield inputs, with source_id)
make_llir  →  compile source   →  fill real dsl_region body  →  inline
```

## Usage (CUDA example)
Full tutorial:

`FlagTree/python/tutorials/tle/raw/cuda/03-matrix-multiplication-smem-defered.py`

Register the DSL with `@dialect` and ` deferred=True` 
```python
from pathlib import Path
from triton.experimental.tle.raw import dialect
import triton.experimental.tle.language.raw as tle_raw

CU_FILE = Path(__file__).parent / "03-matrix-multiplication.cu"

@dialect(name="cuda", file=CU_FILE, extern_func_name="MatMul", deferred=True)
def edsl_deferred(*args, **kwargs):
    ...

@dialect(name="cuda", file=CU_FILE)
def edsl_eager(*args, **kwargs):
    ...
```

Call from the Triton kernel
```python
if DEFERRED:
    acc_smem = tle_raw.call_smem(
        edsl_deferred,
        [acc_smem, a_smem, b_smem],
        output_indices=[0],   # required in deferred mode
    )
else:
    acc_smem = tle_raw.call_smem(
        edsl_eager,
        [acc_smem, a_smem, b_smem],
        # output_indices can be omitted; inferred from LLVM analysis
    )
```


sunnycase (approved):
LGTM

zhzhcookie (approved):
LG

PR #701, opened 2026-06-15
[TLE] Add rank count primitives

Galaxy1458:
## Summary

This PR adds support for querying distributed execution topology information from Triton kernels.

Distributed kernels frequently require access to the local PE rank and the total number of PEs to implement communication patterns such as:

* Point-to-point communication
* Ring-based algorithms
* Reduce-Scatter
* All-Gather
* Distributed synchronization primitives

Providing these queries as first-class TLE operations simplifies distributed kernel development and enables subsequent compiler optimizations.

### Example

```python
local_rank = tle.shard_id(mesh, 'device', comm_ptr=dev_comm_dptr)
n_rank = mesh.shape[0]
peer = (local_rank + 1) % n_rank 

```

This allows Triton/TLE kernels to reason about distributed topology directly on the device side.


zhzhcookie (dismissed):
LGTM

sunnycase (commented):
Thanks for adding these primitives. I have a concern about the user-facing API shape here.

`PE` is not a common term in the GPU kernel programming domain, so exposing it directly in the API may add extra learning cost for users. Also, `device_mesh` already provides a way to query `world_size`, and `tle.shard_id` can be used to get the local id. Given that existing coverage, could you clarify why these additional APIs are necessary and what use cases they enable that the current APIs cannot support?

Galaxy1458:
> Thanks for adding these primitives. I have a concern about the user-facing API shape here.
> 
> `PE` is not a common term in the GPU kernel programming domain, so exposing it directly in the API may add extra learning cost for users. Also, `device_mesh` already provides a way to query `world_size`, and `tle.shard_id` can be used to get the local id. Given that existing coverage, could you clarify why these additional APIs are necessary and what use cases they enable that the current APIs cannot support?

`device_mesh` and `shard_id` are currently configured on the host side, and the `shard_id` API depends on the `device_mesh` configuration. However, in practical training and inference deployments, the number of participating ranks may change dynamically due to scaling, elasticity, or runtime scheduling decisions. In such scenarios, requiring users to preconfigure `device_mesh` introduces additional complexity and operational overhead.

To address this limitation, `n_pes` and `my_pe` as pure device-side primitives. These primitives provide the following advantages:

* **Dynamic runtime discovery**: Device code can directly obtain the current rank identifier (`my_pe`) and the total number of participating processing elements (`n_pes`) at runtime, without relying on host-side configuration.
* **Improved usability**: Users are no longer required to explicitly maintain and synchronize `device_mesh` metadata when the execution topology changes.
* **Alignment with PyTorch distributed execution**: Similar to 
``` torchrun \
        --nproc_per_node=2 \
        --nnodes=1 \
        --node_rank=0 \
```
 where rank-related information is provided by the runtime environment.

sunnycase (approved):
LGTM

zhzhcookie (approved):
LGTM

PR #708, opened 2026-06-18
[BACKEND] Add iluvatar compute spec key specialization

stringl1l1l1l:
## Summary

This PR moves iluvatar `compute_spec_key` specialization behind the FlagTree backend spec mechanism, keeping the default Triton JIT path unchanged while enabling iluvatar-specific 4-byte divisibility keying.

## What’s included
  - add a `spec_func("compute_spec_key")` lookup in JIT binder creation
  - add iluvatar `compute_spec_key` implementation using 4-byte divisibility
  - export the new spec function from `third_party/iluvatar/backend/spec/__init__.py`

## Notes

- This follows the existing FlagTree backend specialization pattern: mainline code exposes only a generic hook, while iluvatar-specific behavior lives under `third_party/iluvatar/backend/spec/...`.


zhzhcookie (approved):
LGTM

PR #718, opened 2026-06-25
[BACKEND] Remove Iluvatar ext_Autotuner_key autotune hook

Salamanca001:
### Summary
Remove the Iluvatar-only ext_Autotuner_key autotune hook and its call site in autotuner.py.

That hook added all int kernel arguments (including philox_offset) to the autotune cache key, so RNG ops such as randn missed the cache on every call and re-ran autotuning (~680ms). Autotune now uses the upstream key logic (key_idx fields + tensor dtypes only).

Test plan

- pre-commit on changed files

- Rebuilt and installed with FLAGTREE_BACKEND=iluvatar

- Verified randn steady-state latency on Iluvatar MR-V100 (~0.15ms vs ~680ms before)

### DIFF
--- a/python/triton/runtime/autotuner.py
+++ b/python/triton/runtime/autotuner.py
@@ -155,9 +155,6 @@ class Autotuner(KernelInterface):
                 if hasattr(arg, "dtype"):
                     key.append(str(arg.dtype))
             key = tuple(key)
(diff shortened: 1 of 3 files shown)


zhzhcookie (approved):
LGTM

PR #732, opened 2026-07-01
[TLE] Add optional offset parameter to tle.remote op

lzllx123:
## Overview
Add an optional scalar `i64` offset parameter to the `tle.remote` op on the triton v3.6.x branch, allowing device-space remote pointers to access peer memory at a specified byte offset.

## Changes
1. Add an optional scalar `i64` offset parameter to `tle.remote`.
2. Restrict the offset to scalar `i64` values only.
3. Use `TT_IntLike` for the `RemotePointersOp` offset operand to align with the verifier.
4. Enforce the offset requirement for device-space remote pointers and reject invalid cases.
5. Fix optional operand validation in the `RemotePointersOp` verifier.
6. Update `test_tle_distributed_d2d.py` to demonstrate the new `offset` parameter in `tle.remote`.


Galaxy1458 (approved):
LGTM

PR #741, opened 2026-07-03
[CI] Update runner label from aipu to aipu3.3

liuhaoyutz:
Update runs-on from 'aipu' to 'aipu3.3' to match the actual label configured on the self-hosted runner machine.


zhzhcookie (approved):
LGTM

PR #745, opened 2026-07-03
[CI] Update runner label from nv-jiuding to nvidia3.5

liuhaoyutz:
Update runs-on from 'nv-jiuding' to 'nvidia3.5' to match the actual label configured on the self-hosted runner machine.


.github/workflows/nv3.5-build-and-test.yml
```diff
@@ -13,8 +13,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  hopper-build-and-test:
-    runs-on: hopper
+  nv3.5-build-and-test:
```
zhzhcookie: '.' is illegal here, so you can’t see the corresponding workflow run.

`nv-build-and-test` is enough.

liuhaoyutz: Fixed and pushed a new patch

.github/workflows/nv3.5-build-and-test.yml
zhzhcookie, line 56: Please delete this code block `FlagTree Build on NVidia (triton_v3.4.x branch)`

liuhaoyutz: Removed v3.5.x support from v3.4.x workflow, pushed a new patch, please help review again.

zhzhcookie (approved):
LGTM

PR #748, opened 2026-07-06
[CI] Update runner label from hopper to nvidia3.5

liuhaoyutz:
Update runs-on from 'hopper' to 'nvidia3.5' to match the actual label configured on the self-hosted runner machine.


zhzhcookie (approved):
LGTM

PR #751, opened 2026-07-06
[TLE] Align and fix do_bench_npu implementation

wangziy1:
* [FEAT] Align do_bench_npu with CUDA's do_bench
  - Add a new ascendC kernel to clear L2 cache and avoid time calculation errors due to torch API limitations
  - Add return mode support to get "min/max/median/mean" statistical results
  - Add matmul kernel before benchmark to stabilize Ascend NPU frequency (cherry picked from commit 63d5c15fed7b182c4e68a544222230ed6dbdcbe7)

* [TLE] Fix sync issues in do_bench_npu
  - Fix synchronisation order when calculating warmup count. The original code called fn() in the estimation loop without synchronisation, which caused event recording to be inaccurate. Added torch.npu.synchronize() after each call to ensure all kernels complete before recording events
  - Move matmul frequency-boosting outside the profiler scope. The MatMul used to raise NPU frequency should run on NPU (not CPU), and must be placed before the profiler context to avoid polluting profile results with irrelevant kernel data (cherry picked from commit a6667467779b402f3c21688af991568b0d9b6ac4)


third_party/ascend/backend/testing.py
```diff
@@ -21,28 +21,28 @@
...
 import triton.runtime as runtime
+from triton._C.clear_l2 import do_bench_clear
 
 
 def get_home_dir():
     return os.getenv("TRITON_HOME", Path.home())
 
 
-def do_bench_npu(funcs, warmup=5, active=30, clear_l2_cache=False, prof_dir=None, keep_res=False, collect_prof=True):
+def do_bench_npu(funcs, warmup=25, active=100, prof_dir=None, clear_l2_cache=False, keep_res=False, collect_prof=True,
```
zhzhcookie, line 36: Please do not change default value and parameter order.

PR #753, opened 2026-07-07
[RPU][CI] Harden board CI: serialize runs, reap all runners, retry transient SSH drops

xup16:
## Problem

The single shared RPU board can only test one thing at a time, and its sshd drops connections under load. Two flaky failure modes were observed on `triton_v3.6.x` CI:

1. **Board-state races.** The per-PR concurrency group let different PRs — and manual re-runs of the same PR — run concurrently/back-to-back on the one board, sharing `~/FlagTree-ci` (rsync `--delete`), the build dir, and `/dev/rpu`. Same source passed in one run and failed the next (run `28788820994`: attempt 3 passed the on-board test at 16:25–16:30 UTC; attempt 6 of the same head `dac38bffea` failed 16:31–16:34, 76s later with nothing concurrent — a residual `launch_kernel_runner` from the prior run, younger than the 130s reap guard, wedged `/dev/rpu`).
2. **Transient SSH kex drops.** The board's sshd drops the connection during key exchange under connection pressure (`kex_exchange_identification: Connection closed`), failing the Sync (rsync) step with `rsync error: code 255`.

## Fix

- **Serialize board runs.** One fixed board-wide concurrency group (`rpu3.6-board`, not per-PR) so only one job drives the board at a time, with `cancel-in-progress: false` (queue) so an in-flight board test is never interrupted mid-run (a cancelled smoke can orphan a runner).
- **Reap all leaked runners.** With runs serialized, no concurrent job owns a runner, so the Test step reaps **all** leaked `launch_kernel_runner` before the smoke instead of only those >130s old (the age guard left the gap above). Match on the real executable via `/proc/PID/exe` so the reap never kills its own shell, whose cmdline also contains `launch_kernel_runner`.
- **Retry transient SSH drops.** The three board-touching steps (rsync, build ssh, test ssh) retry with backoff. For the ssh steps, retry **only on exit 255** (ssh connection error) so a real build/test failure — which returns the remote command's own exit code — is not retried; build and test are idempotent (`rm -rf build`; the smoke reaps first), so a reconnect re-runs safely.

## Testing (on the RPU board)

- **Reap footgun test:** a decoy `bash` shell whose cmdline contains `launch_kernel_runner` (exe = `/usr/bin/bash`) — the old `pgrep -f` flags it (would kill a plain shell), the new `/proc/exe` match correctly ignores it.
- **Full Test step:** reap + `pytest -q third_party/rpu/python/test/unit` = **204 passed, 8 skipped, 0 failed**, then board smoke = **11/11 PASS**.
- `bash -n` clean on all modified run blocks; YAML validated.

The concurrency and retry changes are GitHub-side / driver-side and validated by review.

> Note: the transient-drop root cause was a separate host flooding the board's sshd past its default `MaxStartups 10`; that host has since been firewalled and the board's `MaxStartups` raised to `100:30:200` (board-side, outside this PR). The retry here is defense-in-depth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/rpu3.6-build-and-test.yml
+++ b/.github/workflows/rpu3.6-build-and-test.yml
@@ -7,8 +7,15 @@ on:
     branches: [ "triton_v3.6.x" ]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  # Serialize ALL runs on the single shared RPU board with one fixed, board-wide
+  # group (NOT per-PR): only one job may drive the board at a time. Different PRs
+  # -- and manual re-runs -- otherwise race, sharing ~/FlagTree-ci (rsync
+  # --delete), the build dir, and /dev/rpu, which caused flaky failures. Queue
+  # (cancel-in-progress: false) so an in-flight board test is never interrupted
+  # mid-run (a cancelled smoke can orphan a launch_kernel_runner that wedges
+  # /dev/rpu for the next job).
+  group: rpu3.6-board
+  cancel-in-progress: false
 
 # The RPU CI runner is a small x86 host that orchestrates the build/test on a
 # separate aarch64 RPU board over SSH. The board cannot be reached directly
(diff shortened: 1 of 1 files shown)


zhzhcookie (approved):
LGTM

PR #786, opened 2026-07-13
 Add debugger and CANN profiler support for the Ascend backend

ChijinZ:
This PR adds integrated debugger and CANN profiler support for the Ascend backend.

  It introduces compiler-level instrumentation, runtime data collection, debugger report generation, and CANN-backed Proton profiling. It also adds
  automatic profiler instrumentation so Triton kernels can expose internal execution details in the generated timeline without requiring manual
  instrumentation in user kernels.

  ## Debugger

  - Add debugger instrumentation and metadata MLIR passes.
  - Add runtime buffers and asynchronous export support for Ascend.
  - Add decoding and report generation for:
    - summary records
    - memory accesses
    - full-value references
    - statement-level execution records
  - Add Python APIs for debugger configuration and kernel launch integration.
  - Support level 1 and level 2 data collection.
  - Export text reports, JSON reports, operation logs, raw records, and optional NumPy artifacts.
  - Add two curated regression samples:
    - pointwise `abs`
    - reduction `softmax`

  ## Profiler

  - Add a CANN vendor adapter to Proton.
  - Support CANN runtime, AICore, bandwidth, MSTX, and internal kernel timeline data.
  - Add Proton artifacts for metadata, vendor associations, and Chrome-compatible timelines.
  - Add automatic Triton IR instrumentation for profiler sessions.
  - Preserve and restore existing Triton launch hooks when profiler sessions start and stop.
  - Support explicit CANN device selection through `device_id`.
  - Add host-timing fallback when vendor runtime metrics are unavailable.

  ## Instrumentation Buffer

  The profiler IR record buffer can be configured with:

  ```bash
  export PROTON_IR_RECORD_BUFFER_MB=<size-in-MB>

  The default size is 32 MB.

  The configured capacity is included in the compilation instrumentation mode, so kernels compiled with different buffer capacities use separate cache
  entries.

  ## Build Integration

  - Add debugger libraries and Python bindings to the FlagTree build.
  - Integrate debugger passes into Triton and Ascend compiler pipelines.
  - Integrate the CANN profiler into Proton.
  - Keep debugger support configurable through FLAGTREE_ENABLE_DEBUGGER.
  - Exclude generated kernels, runtime reports, caches, and local build dependencies from Git.

  ## Validation

  The changes were validated in an Ascend CANN container.

  - Ascend wheel build and installation passed.
  - Ruff, YAPF, clang-format, AST, whitespace, and repository hygiene checks passed.
  - Python tests: 47 passed, 2 skipped.
  - Debugger lit tests: 10 passed.
  - Debugger C++ tests: 45 passed.
  - Real debugger abs example passed on Ascend NPU.
  - Real profiler tiny_mlp example passed on Ascend NPU.
  - The generated MLP timeline contained 37,500 events, including internal records for both linear kernels and the ReLU kernel.

  ## Example Outputs

  The profiler produces:

  profile.hatchet
  profile.timeline.json
  profile.meta.json
  profile.vendor.json

  The debugger produces timestamped text and JSON reports with optional operation logs, raw records, and full-value artifacts.


shiptux:
A few thoughts from the packaging side, none of them blocking.

FLAGTREE_ENABLE_DEBUGGER defaults to ON, so every backend eats the build time and .so size for something that's Ascend-only right now. Any reason not to default it OFF and flip it on in the Ascend build?

The debugger and the profiler don't seem to depend on each other. 200 files is a lot to review in one go, so splitting them into two PRs would probably get each one merged faster.

semantic.py, core.py and code_generator.py are on the common path for every backend, so the statement annotation plumbing is there even when the debugger is off. Could that hang off the existing instrumentation_mode knob instead?

On the profiler side, Session.cpp is +847/-90 on top of the new flagtree_profiler/ directory. If the plan is for CANN to come in through Vendor/Adapter.h, does it make sense to land the adapter interface into Proton first and add CANN on top of it? Looks like proton_vendor_adapter_minimal_patch.md is already going that way.

python/triton/compiler/compiler.py
```diff
@@ -485,10 +506,27 @@ def run(self):
         return self._run
 
     def launch_metadata(self, grid, stream, *args):
-        if knobs.runtime.launch_enter_hook is None:
+        debugger_active = False
+        try:
+            debugger_active = importlib.import_module("triton.runtime.debugger").is_active()
```
shiptux, line 511: Runs on every launch, and the bare except hides real import errors here. Could this go through a hook next to launch_enter_hook instead? Same at L356.

ChijinZ:
I’m closing this PR because we plan to move this feature into a separate repository and integrate it into FlagTree as a submodule under `third_party`. This structure will keep the feature decoupled from the main FlagTree repository.

PR #792, opened 2026-07-14
[BUILD] Check cache version instead of destination

liuhaoyutz:
Check the cached binary version in ~/.triton rather than the already-installed binary in third_party/nvidia/backend/. This avoids unnecessary re-downloads when the cache is valid but the destination differs.


zhzhcookie (approved):
LGTM

Galaxy1458 (approved):
LGTM

PR #794, opened 2026-07-14
[BUILD] Restore Debian/RPM packaging lost in history rewrite; add openEuler 24.03 RPM support

shiptux:
## Summary

Restores the Debian/RPM packaging merged in #607 — the merge commit
(`e56773e1`) is no longer an ancestor of `main` after the branch
history was rewritten, so all packaging files vanished from the tree
— and extends it with **openEuler 24.03 LTS** RPM support (part of
the FlagOS-on-openEuler effort; same pattern as
flagos-ai/FlagSparse#29 and flagos-ai/FlagAttention#35, both green).

## What changed

**Recovered from the #607 merge commit (unchanged):**
`packaging/{debian,rpm,INSTALL.md,README.md,spike}`,
`.github/workflows/build-{deb,rpm}.yml`, `.dockerignore`.

**Re-ported to the post-rewrite tree:**

- `setup.py` / `pyproject.toml` / `MANIFEST.in` moved to the repo
  root — wheel-builder COPY lists and build WORKDIR updated in both
  Dockerfiles. The narrow COPY is kept on purpose: packaging-only
  edits must not invalidate the ~50 min wheel layer.
- The new `tileir` default backend requires **GCC >= 13** and the
  `cuda-tile` submodule; distro toolchains can't satisfy that
  (openEuler 24.03 ships GCC 12.3). Added a
  `FLAGTREE_DEFAULT_BACKENDS` env knob in
  `python/setup_tools/utils/tools.py` (one field change, default
  unchanged: `nvidia,amd,tileir`); packaging builds pin
  `FLAGTREE_DEFAULT_BACKENDS=nvidia,amd` — the same contents the
  historical wheel/packages shipped.

**openEuler 24.03 additions:**

- `build-flagtree-rpm.sh`: `RPM_BASE_IMAGE` env passthrough
  (default `fedora:43` unchanged).
- `build-rpm.yml`: `fedora43` + `openeuler2403` matrix
  (`fail-fast: false`); the Fedora artifact keeps its exact name
  (`flagtree-nvidia-rpm-packages`), openEuler uploads as
  `flagtree-nvidia-oe2403-rpm-packages`.
- `Dockerfile.rpm`: reconstructs the empty `%dist` on openEuler
  images (`.oe2403`).
- `flagtree.spec`: `%%`-escaped macros in preamble comments —
  openEuler's rpm expands macros inside comments, and the bare
  `%install` in the usage note terminated the preamble
  ("Version field must be present" errors).

## Tested

[...]

PR #796, opened 2026-07-15
[BUILD] Fix packaging CI: ENOSPC, PR-trigger removal, tag matching, version sync

shiptux:
Follow-up to #794. Fixes the packaging CI issues seen this week (two Build-RPM runs hit ENOSPC on 2026-07-15 and blocked unrelated PRs).

- **Narrow the PR trigger to packaging files** — the old `python/**`-wide filter ran the ~45-min build on unrelated PRs (e.g. an mthreads change building the nvidia package). PRs now trigger only on `packaging/**` / workflow-file changes; source-code coverage moves to release tags + `workflow_dispatch`.
- **Fix ENOSPC** — free ~25-30 GB of unused preinstalled toolchains before building (guarded to GitHub-hosted runners only), and drop the CMake build tree + LLVM cache from the wheel Docker layer.
- **Fix tag trigger** — `v*` never matched: release tags are unprefixed since 0.3.0 (`0.6.0`, `0.6.0rc1`). Match numeric tags and skip backend-variant tags (`0.6.0+mthreads3.6`) at the job level.
- **Sync package version** — spec/changelog said 0.5.0 while the wheel is 0.6.0; bump both and fail the rpm install step on future drift.

## Why ENOSPC: disk breakdown of one packaging job

`ubuntu-latest` runners have ~14-20 GB free (varies per machine — hence the flakiness). Peak usage per job:

| Segment | Size |
|---|---|
| Checkout + buildx context + base/toolchain/pip layers | ~2.5-3 GB |
| **Wheel-build layer (peak)**: LLVM download + extract (~3 GB), CMake tree with `libtriton.so` (~4-8 GB), setuptools' two full staging copies (~2-3 GB), wheel zip ×2 (365 MB each) | **~8-13 GB** |
| RPM assembly (buildroot unpacks the wheel) + output | ~2.5-3.5 GB |

Total peak ~13-18 GB rides exactly on the free-space line; both observed failures happened inside the wheel-build peak (fedora43 while zipping the wheel, openeuler2403 at final link). The cleanup step raises headroom past the peak; the in-layer `rm` stops the dead build tree from being carried through the assembly stage.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -1,22 +1,17 @@
 name: Build Debian Packages
 
+# Packaging is a release artifact, not a PR gate: the full-source wheel
+# build costs ~45 min per run and its ENOSPC flakiness was blocking
+# unrelated PRs. Build on release tags or on demand (workflow_dispatch).
 on:
   push:
     tags:
       - 'v*'
-  pull_request:
-    branches: [main]
-    paths:
-      - 'packaging/debian/**'
-      - 'python/**'
-      - 'include/**'
-      - 'lib/**'
-      - 'cmake/**'
-      - 'bin/**'
-      - 'CMakeLists.txt'
-      - 'pyproject.toml'
-      - '.github/workflows/build-deb.yml'
   workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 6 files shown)


zhzhcookie (approved):
LGTM

PR #800, opened 2026-07-16
[KUNLUNXIN] Xpu/triton 3.6 migration

dongjibin1996:
## Summary
- migrate the Kunlunxin XPU backend to Triton 3.6
- keep XPU autotuner compatibility for `generate_configs`, including FlagGems MM/BMM/ADDMM callers
- update SDNN prebuilt objects for Triton 3.6
- preserve size-one `tt.make_range` on XPU so per-core offsets survive lowering
- align LLVMXPU lowering with the LLVM 22 XPU intrinsic signatures
- add the XPU `tl.atomic_mul` frontend, IR binding, and GM2LM lowering used by FlagGems
- keep masked-load lowering aligned with internal Triton by default; explicit `other` simulation remains opt-in via `TRITONXPU_OTHER_SIM`
- fix the packaged launcher library path to `device/xpu3/liblaunch.a` so the wheel links the correct static launcher
- fall back to the original autotuner metadata when an AABS-adjusted square-tile config fails to compile, restoring the working configuration
- align reduction/scan scratch offset reuse with internal Triton in `ReduceOpHelper` and `ScanLoweringHelper`
- document the isolated `llvm_trust` XPU codegen boundary without disabling vectorization

## Validation
- pre-commit: passed
- clean XPU build from an empty CMake build directory: passed
- clean wheel install/import: passed; backend is `xpu`, `tl.atomic_mul` symbol is available (FP16 remains rejected by FlagTree semantic checks)
- note: the earlier clean build produced wheel SHA256 `75c5bb4825bbadc4c865d24378862b3fb7b62dcbc289af5325617a080ec45b61` and `libtriton.so` SHA256 `75c593b7c9e7b79c0a78dc003aba2db9d49f2c17f365179832e4841b6f598f8d`; the launcher-path, AABS-fallback, and scratch-alignment commits change source, so a rebuilt wheel is required for a current hash
- `test_bmm.py --quick --ref cpu`: 3 passed
- final clean A/B over all 150 historical FlagGems entries in both isolated environments: FlagTree 127 passed / 20 failed / 3 timed out; pristine internal Triton `d3c64dd6` 130 passed / 18 failed / 2 timed out
- 126 entries pass in both environments; 19 shared non-passing entries are baseline limitations, not FlagTree-only regressions
- `rms_norm` and `replication_pad3d` pass the final manifest after the compatibility fixes and are no longer FlagTree-only differences
- full clean-build and 150-run report: `third_party/xpu/docs/triton-3.6-validation.md`
- `test_mm.py::test_mv`: 1 passed, 2 skipped
- `test_mm_int4_wo_int8.py`: 1 passed
- `test_mm_all.py`: 19 passed, 1 skipped
- `test_moe_fc_int4_with_preprocess_balanced.py`: 1 passed
- `test_moe_fc_bf16_with_preprocess_balanced.py`: 2 passed
- `test_moe_fc_fp32_with_preprocess_balanced.py`: 1 passed
- `test_moe_all.py`: 8 passed
- `test_sdnn.py`: 800 passed, 6 skipped; 8 known `tl.gelu` exceptions

## Known limitations (shared baseline unless noted)
- **`llvm_trust` codegen:** large BF16 broadcast `mul` and BF16 `repeat_interleave` pass with pristine internal `xtdk-llvm22` but fail with `llvm_trust`. Same-IR dual-toolchain and crossed `llc`/`elfconv` experiments isolate this to `llvm_trust` XPU LLVM codegen. Vectorization stays enabled; see `third_party/xpu/docs/llvm-codegen-validation.md`.
- **`cross_entropy_loss`:** intermittent backward gradient zero-holes reproduce in both FlagTree and pristine internal Triton, so this is a shared XPU nondeterministic correctness issue, not a cache-hit artifact or FlagTree-only regression. FlagTree/internal TTIR and TTXIR lower the output identically. The root cause is not yet resolved and remains under investigation.
- **`atomic_mul`:** the symbol is available, but both trees lower FP16/FP32 multiply through synchronous `GM2LM -> mul -> LM2GM` without a hardware atomic or CAS loop, so duplicate-address contention is nondeterministic. The FlagTree FP16 frontend difference must not be resolved by only allowing FP16; a valid fix needs a native atomic lowering or a correct CAS loop.
- **`scatter` (FP16 multiply):** the only remaining FlagTree-only difference; it is tied to the shared non-atomic XPU multiply lowering above and must not be separated from it.
- **`grid_sample`:** 2D tiled masked-memory correctness is shared and mitigated with `TRITONXPU_OTHER_SIM=1` + `TRITONXPU_STORE_MASK_SIM=1` (or `TRITONXPU_IS_USE_MASK_ZERO=1` with a new enough XRE). 3D border/reflection fail in the shared `TritonXPUVectorize` pass in both trees. The FlagTree-only full-suite timeout is dominated by AABS dependency-analysis overhead (`FLAGTREE_AABS=0` is diagnostic-only); no correctness regression.
- **AABS square-tile verifier:** adjusted `8x8`/`16x16` configs can fail in `TritonXPULegalize`; the runtime fallback restores the original config and keeps affected operators usable, but the verifier defect remains open.
- **`instance_norm`:** shared small-shape numerical sensitivity under the current strict tolerance; fixed-input and 200-seed deterministic comparisons match between both trees.
- FlagTree main does not expose the internal-only `tl.gelu` primitive.
- Five mixed BF16/FP32 cases fail in the PyTorch XPU reference `torch.mm` before launching Triton; internal Triton reproduces the same behavior.
- Direct invocation of `test_add.py`, `test_layernorm.py`, and `test_dot.py` has unbound pytest fixture parameters and fails during collection.


dongjibin1996:
Added `fad41f3` (`[XPU] Handle non-power-of-two constancy`) to the PR head. The change falls back from elementwise result deduplication when constancy is not a power of two, fixing the XPU `kron` assertion for constancy 6. Validation: clean incremental Triton target build succeeded; FlagGems `tests/test_kron.py --ref cpu --quick -q` reports **90 passed**.

include/triton/Conversion/TritonGPUToLLVM/ElementwiseOpToLLVMBase.h
```diff
@@ -120,7 +95,8 @@ class ElementwiseOpConversionBase : public ConvertOpToLLVMPattern<SourceOp> {
     auto invReg = inv.sublayout(dims, {kReg});
     auto bases_inv = invReg.getBases();
     for (auto [c, d] : llvm::zip(constancy, dims)) {
-      assert(llvm::isPowerOf2_32(c));
+      if (!llvm::isPowerOf2_32(c))
+        return resultVals;
```
zhzhcookie:
```C++
#ifdef __XPU__
      if (!llvm::isPowerOf2_32(c))
        return resultVals;
#endif
      assert(llvm::isPowerOf2_32(c));
```

CMakeLists.txt
```
elseif(FLAGTREE_BACKEND STREQUAL "xpu")
  add_definitions(-D__XPU__)
```

third_party/xpu/CMakeLists.txt
```
    # target_compile_definitions(TritonXPU PRIVATE __XPU__)
```

dongjibin1996: Thanks for the suggestion. Since this fallback is XPU-specific, I plan to move it into an XPU-only whole-file overlay:

  third_party/xpu/backend/spec/include/triton/Conversion/TritonGPUToLLVM/ElementwiseOpToLLVMBase.h

  That overlay is selected only for FLAGTREE_BACKEND=xpu and takes precedence over the shared include tree, while non-XPU builds retain the original assertion behavior. This avoids adding __XPU__ as a
  global compile definition and keeps XPU-specific behavior out of the shared header.

  I will update the overlay provenance and validate it with a fresh XPU build. Would this overlay-based approach be acceptable?

zhzhcookie: Good suggestion! However, please carefully verify that the spec `include_directories` come before the main `include_directories` in CMakeLists.txt.

dongjibin1996: > Good suggestion! However, please carefully verify that the spec `include_directories` come before the main `include_directories` in CMakeLists.txt.

Thanks for the suggestion. I initially moved the non-power-of-two fallback into the XPU overlay and restored the assertion in the shared header. However, the CI result exposed another backend-
specific case: the Enflame GCU400 job passed on fad41f3 but failed with exit code 133 after the shared assertion was restored in 0cd06ea.

I therefore kept the shared assertion unchanged and added the same conservative fallback to the Enflame backend overlay under third_party/enflame/backend/spec/include. This keeps the backend-
specific behavior under third_party, avoids a global __XPU__ or other backend macro, and follows the existing backend spec include precedence.

I also added Enflame provenance documentation and a focused structural test. Locally, both the Enflame and XPU tests pass, and the Enflame ElementwiseOpToLLVM.cpp.o compile command confirms that
the backend spec include path precedes the shared include path. The follow-up commit is 5896057.

dongjibin1996: Thanks. I have updated this without introducing a global `__XPU__` definition.

- The shared header and the top-level `CMakeLists.txt` are now identical to the PR base.
- The XPU-only fallback remains in `third_party/xpu/backend/spec/include/.../ElementwiseOpToLLVMBase.h`, with the spec include directory taking precedence.
- The four whole-file XPU source overrides are explicitly wired in `third_party/xpu/CMakeLists.txt` using target-scoped `HEADER_FILE_ONLY` plus `target_sources`.
- `${ARG_UNPARSED_ARGUMENTS}` in the shared `add_triton_object` helper remains unchanged.

I verified the source selection in a fresh Ninja graph, completed a full XPU build and the README-recommended pip installation, and ran the focused structural and XPU FlagGems tests.

zhzhcookie (approved):
LGTM

PR #833, opened 2026-07-22
[CI] Refine CI test trigger conditions

liuhaoyutz:
Replace file filtering with explicit NEED_CI_TEST white list CI triggers only for: workflow itself, white-listed files/dirs, or backend-specific changes.


zhzhcookie (approved):
LGTM

PR #835, opened 2026-07-23
[ENFLAME][TLE-Raw] Support tle raw in defer mode and tle pipe

baoqiliu:
- [common] Add dsl_file_name and extern_func_name parameters to the deferred
  TLE Raw region creation path (C++ & Python bindings), storing them
  as tle_raw.dsl_file_name / tle_raw.extern_func_name attributes on
  the DSLRegionOp for downstream compilation passes.

- [common] Introduce cache_key property on TOPSJITFunction, computed from the
  .tops source content, architecture, and extern_func_name.  Hook it
  into DependenciesFinder (jit.py) so that editing the .tops file
  correctly invalidates the Triton compile cache.

- [enflame] Add register_pending_source() and create_region_deferred() to the
  TOPSJITFunction runtime for the deferred compilation pipeline.

- [enflame] Reformat gcu_dialect.py and mlir_runtime.py for line-length
  compliance (no semantic changes).

- [enflame] update enflame backend to support tle raw in defer mode and tle pipe


python/triton/runtime/jit.py
```diff
@@ -157,6 +157,13 @@ def record_reference(self, val, var_dict=None, name=None):
             return
 
         if getattr(val, "__triton_builtin__", False):
+            # TLE-Raw dialect objects (e.g. TOPSJITFunction, CUDAJITFunction,
+            # MLIRJITFunction) carry a cache_key that hashes their external
+            # source file content.  Fold it into the dependency hash so that
+            # editing the .tops/.cu/.mlir file invalidates the compile cache.
+            dialect_cache_key = getattr(val, "cache_key", None)
+            if dialect_cache_key is not None:
```
i3wanna2: We would prefer not to add another cache-key path in jit.py here. Rather, we already fold TLE-Raw source dependencies into DependenciesFinder via __triton_tle_raw_source_cache_key__; if you need a custom key (e.g. to include arch), please set that attribute on TOPSJITFunction as a string or zero-arg callable—bind_tle_raw_source_cache_key will not overwrite it.

baoqiliu: got, thx~ we would follow zy's plan.

i3wanna2 (approved):
LGTM

Galaxy1458 (approved):
LGTM

PR #837, opened 2026-07-23
[TLE] Fix helper inlining for warp-specialized local pointers

### LINKED ISSUE #828: [BUG] SMEM atomic (atom.shared) deadlocks inside a warp_specialize worker partition
## Summary

A shared-memory atomic (`tl.atomic_add(local_ptr(smem, ...), 1, scope="cta")`, which lowers to `atom.shared`) **hangs the kernel forever when it is placed inside a `tle.gpu.warp_specialize` worker partition**. The exact same atomic runs correctly in a plain (non-warp-specialized) kernel.

## Environment

| | |
|---|---|
| Repo | flagos-ai/FlagTree |
| Commit | `ac3948a9` (branch [`pr787`](https://github.com/flagos-ai/FlagTree/pull/787)) |
| triton | 3.6.0 |
| GPU | NVIDIA H100 80GB HBM3 |
| Driver / CUDA | 595.71.05 / 12.8 |

Also reproduces on [PR#682](https://github.com/flagos-ai/FlagTree/pull/682).

## Minimal reproduction

```python
import triton
import triton.language as tl
import torch

NEXP = 16

@triton.jit
def _noop(o):
    pass

# The atomic body. Fixed slot 0, no index arithmetic, no mask, no input used.
@triton.jit
def _body(out, NEXP: tl.constexpr):
    smem = tle.gpu.alloc([NEXP], dtype=tl.int32, layout=None,
    vec = tle.gpu.local_ptr(smem, (tl.arange(0, NEXP),))
    tl.store(vec, tl.zeros([NEXP], dtype=tl.int32))
    tl.atomic_add(tle.gpu.local_ptr(smem, (0,)), 1, sem="relaxed", scope="cta")
    tl.store(out + tl.arange(0, NEXP), tl.load(vec))

@triton.jit
def plain(out, NEXP: tl.constexpr):
    _body(out, NEXP)

@triton.jit
def ws(out, NEXP: tl.constexpr):
    tle.gpu.warp_specialize([(_noop, (out,)), (_body, (out, NEXP))], [4], [232])

out = torch.zeros(NEXP, dtype=torch.int32, device="cuda")

# WORKS:
plain[(1,)](out, NEXP, num_warps=4); torch.cuda.synchronize()
print("plain ok")

# HANGS FOREVER (no error, no progress):
ws[(1,)](out, NEXP, num_warps=4); torch.cuda.synchronize()
print("ws ok  <- never reached")
```

`plain` prints `plain ok`. `ws` never returns — `cuda.synchronize()` blocks indefinitely.

## Expected

The atomic should behave the same inside a warp_specialize worker as it does in a plain kernel (or, if unsupported, fail at compile time — not hang at runtime).

sunnycase:
## Summary

This change enables the normal Triton/MLIR inliner for operations owned by the TLE dialect. It fixes the shared-memory atomic deadlock reported in #828 by allowing a `@triton.jit` worker containing `tle.local_pointers` to be inlined into its `tle.gpu.warp_specialize` partition before barrier analysis and lowering.

The change:

- registers `TritonInlinerInterface` on `TleDialect`;
- adds regression coverage for ordinary and warp-specialized helpers containing `tle.local_pointers`;
- verifies that explicit `noinline = true` is still respected;
- covers `tle.dsl_region`/`tle.yield`, pipe lifecycle operations, and another effectful TLE operation so the interface is not accidentally limited to `tle.local_pointers`.

Fixes #828.

## Background and root cause

A warp-specialized worker is initially represented as a normal `tt.call` to a private `@triton.jit` helper. The common inliner runs before TLE local pointers are lowered, and every operation in the callee must be legal to inline.

TLE did not register a dialect inliner interface. Consequently, a helper containing `tle.local_pointers` was considered illegal to inline even when the function had `noinline = false`. Helpers containing only operations from dialects that already expose inliner legality could still be inlined, which made the failure appear specific to the atomic/local-pointer worker.

Leaving the call in place is unsafe for this warp-specialized case:

1. shared-memory atomic analysis inserts a CTA barrier in the out-of-line helper;
2. the warp-specialization barrier rewrite only sees barriers lexically contained in the worker partition and therefore cannot assign the helper barrier to the worker barrier group;
3. the generated PTX contains a call from the worker partition to a helper with a bare `bar.sync 0`;
4. only the worker's 128 threads enter that helper while the CTA contains 256 threads, so the barrier waits forever.

The plain kernel does not expose the same mismatch because the helper is entered by the full CTA.

## Fix

`TleDialect` now registers the existing `TritonInlinerInterface`. This marks TLE operations as legal to clone through the standard module inliner and uses Triton's existing call/return handling.

After the change, the worker helper is inlined early, so later shared-memory and warp-specialization passes see the atomic and its barriers directly inside the partition. The resulting PTX has no helper call and uses worker-scoped barriers (`bar.sync 2, 128`) instead of `bar.sync 0`.

This is intentionally a dialect-level fix rather than a special case for `tle.local_pointers`. The current TLE dialect has no custom callable, call, or symbol operations. Its only region/terminator pair is `tle.dsl_region`/`tle.yield`, which is cloned as part of the containing operation and is covered by the new test.

The existing `inlinePipeHelperCalls` code in `TleLowerPipeToNvws` addresses a related symptom but is narrower: it force-inlines only helpers transitively containing pipe lifecycle operations, runs immediately before pipe lowering, and supports only single-block callees. It remains useful as a pipe-specific fallback; this PR fixes the missing general inliner legality that affects local pointers and other TLE operations.

## Validation

- `triton-opt third_party/tle/test/GPU/test_tle_inline.mlir -split-input-file -inline | FileCheck third_party/tle/test/GPU/test_tle_inline.mlir`
- filtered TLE lit test: 1 passed
- `python/test/tle/unit/test_tle_gpu_local_ptr.py`: 12 passed
- `python/test/tle/unit/test_tle_wgmma_pipeline_routing.py`: 6 passed
- existing TritonGPU and Gluon inliner tests passed
- the #828 reproducer completed successfully on an NVIDIA H800; synchronization returned and the output was `[1, 0, ..., 0]`
- generated PTX: no worker helper call, no bare `bar.sync 0`, and two `bar.sync 2, 128` instructions

A full TLE lit invocation is currently obstructed by pre-existing C-style license headers in existing `.mlir` tests, which the current MLIR parser rejects at line 1. The new test uses parser-compatible line comments, and the two existing pipe-lowering tests pass when run without those headers.


Galaxy1458 (approved):
LGTM

zhzhcookie (approved):
LGTM

PR #844, opened 2026-07-24
[TLE] Stabilize FP16 cumsum correctness test

sunnycase:
## Background

`test_tle_cumsum_exclusive_and_total` uses random normal FP16 inputs and compares the result of Triton's parallel scan against a PyTorch reference with fixed tolerances. This made the test intermittently fail in local and CI runs even though repeated launches with the same input were deterministic.

## Root cause

Both `tl.cumsum` and `tle.cumsum` keep FP16 inputs in FP16 during accumulation. Their parallel scan tree can round differently from the PyTorch reference. Depending on the randomly generated input, the difference can slightly exceed the fixed `atol=2e-2, rtol=2e-2` threshold.

A sweep over seeds 0 through 5000 reproduced the failure for 64 of 5001 inputs. For every failing seed, `tle.cumsum` and native `tl.cumsum` produced bit-identical results, confirming that this was a test stability issue rather than a TLE-specific race or lowering error.

## Fix

Generate bounded integer-valued inputs for the FP16 correctness case and cast them to FP16. With values in `[-2, 2]`, every prefix sum in the tested block remains exactly representable in FP16. This preserves coverage of the FP16 lowering path while making the correctness assertion independent of parallel scan accumulation order.

FP32 and BF16 cases continue to use random normal inputs.

## Validation

- `python -m pytest -q python/test/tle/unit/test_tle_cumsum.py`: 12 passed, 1 skipped
- FP16 seed sweep over 10,000 inputs: 0 failures, including exact exclusive and total comparisons


zhzhcookie (dismissed):
LGTM

Galaxy1458 (dismissed):
LGTM

i3wanna2 (dismissed):
LGTM

zhzhcookie (approved):
LGTM

Galaxy1458 (approved):
LGTM

PR #852, opened 2026-07-27
[CI] Add TLE label for TLE-related files

liuhaoyutz:
Add a new TLE label rule to automatically label PRs that modify files related to the TLE component. The rule matches:
- Files under tle/TLE/Tle directories
- Files whose path contains tle/TLE/Tle Note: Only matches exact case patterns 'tle', 'TLE', 'Tle'. Other combinations like 'tLe', 'tlE', 'tLE' are not matched.


zhzhcookie (approved):
LGTM

Issue #846, opened 2026-07-24
[Bug][TLE-Raw][nvshmem] NVSHMEM JIT post-compile hook incorrectly applies nvshmemx_cumodule_init to plain Triton kernels


Fucov:
## Description

When the NVSHMEM TLE Raw CUDA dialect is registered, a plain Triton kernel that does **not** contain any NVSHMEM calls may fail during its first JIT compilation with:

```text
Cuda failure 'named symbol not found'
AssertionError: nvshmemx_cumodule_init failed: 7
```

The issue appears to be related to the NVSHMEM `jit_post_compile_hook` installed by TLE Raw.

After registering an NVSHMEM dialect, the hook seems to be applied globally to subsequently compiled Triton CUDA modules. As a result, `nvshmemx_cumodule_init()` is also invoked for ordinary Triton kernels whose modules do not contain the NVSHMEM device symbols expected by NVSHMEM.

A minimal reproducer is provided below.

---

## Environment

```text
GPU: NVIDIA H20, 2 GPUs used
FlagTree commit: <COMMIT_HASH>

Python: 3.12
CUDA Runtime: 13.0
CUDA Driver: 13.2

NVSHMEM: 3.4.5
NCCL: 2.28.9+cuda13.0
```

Relevant NVSHMEM initialization output:

```text
NVSHMEM v3.4.5
NVSHMEM configuration:
  CUDA API                     13000
  CUDA Runtime                 13000
  CUDA Driver                  13020
  NVSHMEM_NVTX=ON
  NVSHMEM_USE_NCCL=ON
```

Both NVSHMEM ranks initialize successfully before the failure.

---

## Minimal Reproducer

The following program does two things:

1. Registers an NVSHMEM TLE Raw dialect.
2. JIT-compiles a completely ordinary Triton `add` kernel.

The Triton kernel itself contains no `tle_raw.call()` and no NVSHMEM operation.

```python
import os
from pathlib import Path

import torch
import torch.distributed as dist
import triton
import triton.language as tl

from triton.experimental.tle.raw import dialect
from triton.experimental.tle.raw.nvshmem.utils import (
    init_nvshmem_by_torch_pg,
    init_torch_distributed,
    load_common_host,
)


# Registering this dialect installs the NVSHMEM CUDA runtime hook.
# The function is intentionally never called by the Triton kernel below.
@dialect(
    name="cuda",
    library="nvshmem",
    compiler="clang",
[...]

PR #857, opened 2026-07-28
[KMCompiler][TLERaw] Fix incorrect injection of NVSHMEM cumodule_hook

lizhangyu258:
This PR fixes incorrect hook injection in the following cases:
1. A single Python file launches multiple kernels, but only one of them uses NVSHMEM;
2. A CUDA device function is registered in Python, but it is never invoked by the Triton kernel through `tle_raw.call`, https://github.com/flagos-ai/FlagTree/issues/846;

Our solutions: 
- For case1, bind the hook to the kernel and write it into the kernel metadata; 
- For case2, defer the hook registration until inside `tle_raw.call`;

### DIFF
--- a/python/triton/experimental/tle/language/raw/core.py
+++ b/python/triton/experimental/tle/language/raw/core.py
@@ -60,15 +60,10 @@ def _normalize_hint(hint):
     return str(hint) if hint else ""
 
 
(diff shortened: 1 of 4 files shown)


i3wanna2 (approved):
LGTM

zhzhcookie (approved):
LGTM

PR #845, opened 2026-07-24
Revert "[TLE] Fix helper inlining for warp-specialized local pointers"

zhzhcookie:
Reverts flagos-ai/FlagTree#837

PR #859, opened 2026-07-28
[TLE] Reland scoped helper inlining for warp-specialized local pointers

### LINKED ISSUE #828: [BUG] SMEM atomic (atom.shared) deadlocks inside a warp_specialize worker partition

sunnycase:
## Summary

Relands #837, reverted by #845, with narrower TLE-specific inlining legality.

Helpers containing `tle.local_pointers` were rejected by the module inliner. In the warp-specialized case from #828, leaving the worker helper out of line allowed barrier insertion to produce a CTA-wide barrier that only worker threads entered, causing a deadlock.

This version:

- registers a dedicated `TleInlinerInterface` instead of reusing `TritonInlinerInterface`;
- allows only regionless, non-terminator TLE operations to be cloned with an enclosing Triton function;
- keeps `tle.dsl_region` / `tle.yield` and calls inside isolated TLE regions opaque to generic inlining;
- continues to respect `noinline = true`.

This avoids inheriting Triton's `tt.return`-specific terminator handling and unconditional region legality while restoring the fix for #828.

## Root cause and impact

The common inliner runs before TLE local pointers are lowered, and every operation in the callee must be legal to inline. TLE previously had no dialect inliner interface, so a helper containing TLE operations remained as a `tt.call`.

For a warp-specialized worker, shared-memory atomic analysis could then insert a CTA-wide barrier inside that out-of-line helper. Only the worker's 128 threads entered the helper in a 256-thread CTA, so the generated bare `bar.sync 0` could wait forever. Inlining the helper early lets the warp-specialization passes see and scope those barriers correctly.

## Validation

- TLE inliner MLIR regression with `-split-input-file -inline -verify-each` and FileCheck;
- 5 adjacent TritonGPU, TritonNvidiaGPU, Gluon, membar, and warp-specialization MLIR regressions;
- TLE cumsum, local-pointer, and WGMMA routing unit tests: 30 passed, 1 skipped;
- #828 H800 reproducer: plain and warp-specialized kernels completed with the expected output;
- generated PTX contains no helper call or bare `bar.sync 0`, and contains two worker-scoped `bar.sync 2, 128` instructions.

Fixes #828.

### DIFF
--- a/third_party/tle/dialect/lib/IR/Dialect.cpp
+++ b/third_party/tle/dialect/lib/IR/Dialect.cpp
@@ -23,8 +23,8 @@
 
 #include "tle/dialect/include/IR/Dialect.h"
 #include "mlir/Support/LLVM.h"
-#include "triton/Dialect/Triton/IR/Interfaces.h"
 #include "tle/dialect/include/IR/Dialect.cpp.inc"
+#include "triton/Dialect/Triton/IR/Interfaces.h"
 
 #define GET_ATTRDEF_CLASSES
 #include "tle/dialect/include/IR/TleAttrDefs.cpp.inc"
--- a/third_party/tle/dialect/lib/IR/Dialect.cpp
+++ b/third_party/tle/dialect/lib/IR/Dialect.cpp
@@ -23,8 +23,8 @@
 
 #include "tle/dialect/include/IR/Dialect.h"
 #include "mlir/Support/LLVM.h"
+#include "mlir/Transforms/InliningUtils.h"
 #include "tle/dialect/include/IR/Dialect.cpp.inc"
-#include "triton/Dialect/Triton/IR/Interfaces.h"
 
 #define GET_ATTRDEF_CLASSES
 #include "tle/dialect/include/IR/TleAttrDefs.cpp.inc"
@@ -38,6 +38,18 @@
 #endif
 
 namespace mlir::triton::tle {
+namespace {
+struct TleInlinerInterface final : public DialectInlinerInterface {
+  using DialectInlinerInterface::DialectInlinerInterface;
+
+  bool isLegalToInline(Operation *op, Region *, bool, IRMapping &) const final {
+    // TLE regions and terminators have their own semantics. Only allow
+    // ordinary operations to be cloned with a containing Triton function.
+    return op->getNumRegions() == 0 && !op->hasTrait<OpTrait::IsTerminator>();
+  }
+};
+} // namespace
+
 void TleDialect::initialize() {
   addAttributes<
 #define GET_ATTRDEF_LIST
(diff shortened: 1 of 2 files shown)


Galaxy1458 (approved):
LGTM

i3wanna2 (approved):
LGTM

zhzhcookie (approved):
LGTM

PR #871, opened 2026-07-29
[CI] Add clear-cache-if-labeled action

liuhaoyutz:
Add a composite action that checks if a PR has a specific backend ClearCache label (e.g., ClearCache:mthreads) and clears cache if present.


zhzhcookie (approved):
LGTM

PR #876, opened 2026-07-30
[CI] Fetch PR labels via API in clear-cache action

liuhaoyutz:
Use GitHub REST API instead of static event payload to fetch PR labels. This ensures that manually adding labels and re-running the job will successfully trigger the cache cleanup.


zhzhcookie (approved):
LGTM

PR #881, opened 2026-07-30
[TLE] Add tle.concat_dot_fragments to merge dot operands along K

zeroherolin:
## Background

A quantized MoE kernel dequantizes its weight operand inside the main loop. One `tl.inline_asm_elementwise` unpacks an int32 of packed 4-bit weights into eight bf16 outputs, and because `tl.dot` hides the fragment layout it picks for the mma, the host has no way to learn the lane→element mapping and must pack with a strided permutation to accommodate it. The dequant therefore yields eight K-slices that are contiguous in K but are eight independent tensors, and `tl.dot` accepts a single tensor per operand — so the loop issues one mma per slice and repeats the address computation, the token load and the fence/wait every time.

Assembling the slices in the kernel does not help. `tl.join` + `reshape` produces a blocked layout, which has to be converted back to `dot_op` before it can feed the mma — materializing as a `local_alloc` / `local_load` pair, i.e. a shared-memory round trip of the dequantized weight on every K iteration.

The merge should be free. In the LinearLayout algebra a `dot_op` layout of K=128 is exactly eight K=16 layouts concatenated along K: it adds three register bases (`[0,16]`, `[0,32]`, `[0,64]`) and leaves the lane and warp bases bit-identical, so every element of the wider fragment already lives on the thread that holds it in the slice. What is missing is an op expressing that relabel: `tl.cat` is unordered, and the join/reshape chain changes the layout.

**Motivation**: add a TLE primitive for the K-wise concatenation of dot-operand fragments, so one mma covers the whole K tile — without touching host packing or the dequant asm.

### `tle.concat_dot_fragments(tiles, dim)`

Takes N same-encoding dot-operand fragments and returns their concatenation along `dim`, lowering to a per-thread register gather: for each result register slot the authoritative `LinearLayout` of result and tile tells which `(tile, tile-register-slot)` feeds it, and that register is copied over. No shared memory, no cross-lane traffic, no extra instruction. The hardware still issues the same number of mma instructions; what goes away is the per-segment overhead.

The op is wired through all three layers a user-visible TLE op needs: ODS definition and verifier, a TT→TTG legalization pattern (Triton-level tensors carry no layout encoding, so the result inherits `tiles[0]`'s once TTG assigns one), and the TTG→LLVM lowering. Unlike `extract_tile`, the lowering does not use CTA-tile index arithmetic — that only handles `BlockedEncoding` — but goes through `LinearLayout`, which covers dot-operand encodings.

### Reaching through the op during layout propagation

For the relabel to pay off, the `dot_op` encoding has to reach the dequant through the op; otherwise the operand stays `#blocked` and the round trip returns. The gate is the `noDataMovement` allowlist in `hoistConvertDotOperand`, which treats any op outside it as a leaf of the backward slice — before `inferSrcEncoding` is ever consulted — so the new op is added there.

The accompanying `inferSrcEncoding` / `inferDstEncoding` overloads propagate `DotOperandEncodingAttr` **only**: it carries no K extent and can be shared unchanged between a slice and the merged result, whereas shape-coupled encodings like blocked would yield the wrong per-thread element count. Those return empty and take the generic shared-memory conversion path.

### Constraints, split by what breaks when violated

The relabel derives its register mapping on lane 0 and applies it to every thread. Violating that premise produces silently wrong registers rather than an error, so each premise is checked explicitly.

The **verifier** checks what the types alone determine: at least two tiles, tiles agreeing in shape/dtype/encoding, `dim` in range, and the result matching the tiles in every dim except `dim`, which grows by N.

The **lowering** checks the dot_op specific premises, which cannot live in the verifier: operands only pick up their `dot_op` encoding after layout propagation, so until then a `#blocked` tile is a legal intermediate state.

| Checked when lowering | Why |
|---|---|
| encoding is `dot_op` | other layouts place lanes differently as `dim` grows |
| `dim` is the contraction axis for the operand's `opIdx` | extending M or N changes which lane owns an element |
| tile spans at least `8 * kWidth` along `dim` | below that the layout stops scaling with K, so the tile is not a K-slice of the wider fragment |
| tile and result agree on lane/warp/block bases | otherwise the lane-0 mapping cannot be applied to other threads |
| neither broadcasts along an indexed dim | a broadcast maps several registers to one element, and the reverse lookup would drop some and duplicate others |

Two further conditions need no code: non-power-of-2 tile counts are already rejected by the upstream TTG layout verifier with an accurate diagnostic, and a bijective mapping cannot be required because `dot_op` layouts are inherently broadcast — `broadcastedDotOperandLayout` uses `zeros1D` for lane and warp along K so the A operand can be shared across warps.

Whether the merge is also *profitable* is left to the caller. Canonical register-base order, `local_alloc` count, surviving `convert_layout` and whether the merged mma was split back along K are all per-(shape, config) properties that the op cannot judge, and none of them breaks correctness — the compiler has a legal fallback. Asserting on them would turn "runs but slower" into "does not run".

## Test

- `test_tle_concat_dot_fragments.mlir` — the lowering emits `extractvalue`/`insertvalue` only: 128 in, 128 out for a 2×K=64 merge, and no load, store, barrier or mma anywhere in the function. Covers a batched (rank-3) fragment too.
- `test_tle_concat_dot_fragments_errors.mlir` — seven negative cases, covering the verifier's arity/shape/dim checks and the lowering's encoding, contraction-axis, minimum-K and broadcast checks.
- `test_concat_dot_fragments.py` — on GPU, a merged dot is compared against accumulating one dot per K segment and is bit-identical; a second test reads the emitted TTGIR to confirm the concat kept its `dot_op` encoding and that exactly one mma was issued.

## Performance

Hardware **H20-3e**, bf16, CUDA Graph ON, `return_mode=median`, `USE_FLAGTUNE=0`. Consumer is the FlagGems MXFP4 fused-Marlin-MoE kernel (`fused_marlin_moe.py`, gated by `FLAGGEMS_MXFP4_USE_CONCAT_TILE`); baseline is the same kernel issuing one mma per K segment. `speedup = T_vLLM(Marlin) / T_Gems`.

Shapes: 53 deduped shapes from a real DeepSeek-V4-Flash inference trace, geometry `(M, E=256, H=4096, I=256, top_k=6)`, `M ∈ [1, 16384]`. Both sides autotune independently with the persistent config DB cleared beforehand.

| | segmented | merged |
|---|---|---|
| Average speedup | 1.062 | **1.168** |
| Call-weighted speedup | 1.110 | **1.208** |

Latency improves on **53/53 shapes**, +2.53% to +15.72% (mean +9.33%), no regression. By regime: decode (`M ≤ 128`, 19 shapes) +10.17%, mid (`128 < M ≤ 512`, 32 shapes) +9.00%, prefill (`M > 512`) +6.75%.

The weight GEMM merges 8 segments into one mma and the SiLU-fused gate/up GEMM merges 16 into two. TTGIR at M=256 confirms where the time goes: the per-segment 16-wide activation buffers collapse into one 128-wide buffer (`local_alloc` 8→1, 13→6 in total), and the address computation and async copies that were duplicated per segment drop by roughly half (`async_copy_global_to_local` 65→30, `tt.addptr` 78→41). The dequant inline asm is byte-identical on both sides — the merge does not touch it.

The primitive sits between the dequant output and the mma and is independent of the mma variant, so both autotuner winners benefit — MMAv2 (`tt.dot`) and wgmma (`ttng.warp_group_dot`) alike.

### DIFF
--- a/lib/Conversion/TritonToTritonGPU/TritonToTritonGPUPass.cpp
+++ b/lib/Conversion/TritonToTritonGPU/TritonToTritonGPUPass.cpp
@@ -965,25 +965,57 @@ public:
   }
 };
 
+// concat_dot_fragments op pattern
+class TleConcatDotFragmentsOpPattern
+    : public OpConversionPattern<tle::ConcatDotFragmentsOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(tle::ConcatDotFragmentsOp op,
+                  tle::ConcatDotFragmentsOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto tiles = adaptor.getTiles();
+    if (tiles.empty())
+      return op.emitError("concat_dot_fragments requires at least one tile");
+
+    auto tile0Type = dyn_cast<RankedTensorType>(tiles[0].getType());
+    if (!tile0Type)
+      return op.emitError("tiles must be ranked tensors");
+    auto tileEnc = tile0Type.getEncoding();
+    if (!tileEnc)
+      return op.emitError("tiles must have encoding attribute");
+
+    Type retType = op.getType().cloneWithEncoding(tileEnc);
+
+    auto newOp = rewriter.replaceOpWithNewOp<tle::ConcatDotFragmentsOp>(
+        op, retType, tiles, op.getDim());
+
+    addNamedAttrs(newOp, adaptor.getAttributes());
+
+    return success();
+  }
+};
+
 // flagtree tle raw
 void populateTleRawPatterns(TritonGPUTypeConverter &typeConverter,
                             RewritePatternSet &patterns) {
   MLIRContext *context = patterns.getContext();
-  patterns
-      .add<TleDSLRegionOpPattern, TleExtractTileOpPattern,
-           TleInsertTileOpPattern, GenericOpPattern<tle::LocalPointersOp>,
-           GenericOpPattern<tle::RemotePointersOp>,
-           GenericOpPattern<tle::ExclusiveCumsumOp>,
-           GenericOpPattern<tle::WGMMAOp>, GenericOpPattern<tle::WGMMAWaitOp>,
-           GenericOpPattern<tle::DistributedBarrierOp>,
-           GenericOpPattern<tle::YieldOp>,
-           GenericOpPattern<tle::ExtractAllocatedPtrOp>,
-           GenericOpPattern<tle::ExtractAlignedPtrOp>,
-           GenericOpPattern<tle::ExtractOffsetOp>,
-           GenericOpPattern<tle::ExtractSizesOp>,
-           GenericOpPattern<tle::ExtractStridesOp>,
-           GenericOpPattern<tle::ExtractPtrOp>, GenericOpPattern<tle::PackOp>>(
-          typeConverter, context);
+  patterns.add<
+      TleDSLRegionOpPattern, TleExtractTileOpPattern, TleInsertTileOpPattern,
+      TleConcatDotFragmentsOpPattern, GenericOpPattern<tle::LocalPointersOp>,
+      GenericOpPattern<tle::RemotePointersOp>,
+      GenericOpPattern<tle::ExclusiveCumsumOp>, GenericOpPattern<tle::WGMMAOp>,
+      GenericOpPattern<tle::WGMMAWaitOp>,
+      GenericOpPattern<tle::DistributedBarrierOp>,
+      GenericOpPattern<tle::YieldOp>,
+      GenericOpPattern<tle::ExtractAllocatedPtrOp>,
+      GenericOpPattern<tle::ExtractAlignedPtrOp>,
+      GenericOpPattern<tle::ExtractOffsetOp>,
+      GenericOpPattern<tle::ExtractSizesOp>,
+      GenericOpPattern<tle::ExtractStridesOp>,
+      GenericOpPattern<tle::ExtractPtrOp>, GenericOpPattern<tle::PackOp>>(
+      typeConverter, context);
 }
 #endif
 
--- a/lib/Dialect/TritonGPU/Transforms/RemoveLayoutConversions.cpp
+++ b/lib/Dialect/TritonGPU/Transforms/RemoveLayoutConversions.cpp
@@ -51,6 +51,7 @@
 #endif // __FLAGTREE_RLC_ENHANCE__
 #include "triton/Analysis/Utility.h"
 #ifdef __TLE__
+#include "tle/dialect/include/IR/Dialect.h"
 #include "tle/dialect/include/Transforms/TransformAttrs.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
 #endif // __TLE__
@@ -4142,6 +4143,9 @@ void LayoutRematerialization::hoistConvertDotOperand(
     return (op->hasTrait<OpTrait::Elementwise>() && isMemoryEffectFree(op)) ||
            isa<BroadcastOp, Fp4ToFpOp, ConvertLayoutOp, UpcastFpOpInterface>(
                op) ||
+#ifdef __TLE__
+           isa<triton::tle::ConcatDotFragmentsOp>(op) ||
+#endif
            isView(op);
   };
   // Stop the slice as soon as we find an operation that cannot be done without
--- a/lib/Dialect/TritonGPU/Transforms/Utility.cpp
+++ b/lib/Dialect/TritonGPU/Transforms/Utility.cpp
@@ -434,6 +434,18 @@ static Attribute inferSrcEncoding(GatherOp op, Attribute dstEnc) {
   return dstEnc;
 }
 
+#ifdef __TLE__
+static Attribute inferSrcEncoding(triton::tle::ConcatDotFragmentsOp op,
+                                  Attribute dstEnc) {
+  // Only DotOperandEncodingAttr carries no K extent, so it is the only encoding
+  // a concat can share unchanged with its tiles; shape-coupled ones like
+  // blocked take the generic conversion path.
+  if (isa<ttg::DotOperandEncodingAttr>(dstEnc))
+    return dstEnc;
+  return {};
+}
+#endif
+
 static Attribute inferTransOpDstEncoding(Attribute srcEnc,
                                          ArrayRef<int64_t> shape,
                                          ArrayRef<int32_t> order) {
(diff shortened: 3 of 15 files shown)


zhzhcookie (approved):
LGTM

sunnycase (commented):
Thanks for putting this together. The LinearLayout-based lowering and the performance results look promising.

However, I would prefer not to introduce a new Python API for this. `concat_dot_fragments` is highly dependent on the selected backend layout: it is only valid for dot operands, along the contraction axis, and only when the layout can be implemented as a thread-local register relabel. Exposing it at the Python level means a source-level-valid call may fail much later if layout propagation does not produce a compatible `dot_op` encoding or when compiling for another backend.

I think this should instead be modeled as a backend optimization:

1. Express the ordered concatenation using existing operations, such as the exact `join -> permute -> reshape` pattern.
2. After the MMA layout has been selected, recognize this pattern when its result feeds a dot operand.
3. Verify the contraction axis, shapes, element types, and LinearLayout compatibility, including that the mapping is register-local and bijective.
4. If all checks pass, replace it with an internal `concat_dot_fragments`/TTG op and reuse the register-gather lowering from this PR.
5. If any check fails, leave the original IR unchanged so that the existing lowering remains a correct fallback.

Ideally, the matcher should run after MMA layout selection but before dot operands are materialized into shared memory and before the second remove-layout-conversions pass. It may also be worth checking whether the existing layout propagation can already eliminate the `join/permute/reshape` materialization for some configurations, and whether the reported shared-memory round trip is specific to enhanced RLC, MMAv2, or the full consumer IR.

I would not recommend starting by automatically merging the current chain of eight `tl.dot` operations. That transformation also needs to prove floating-point accumulation legality and merge the segmented activation loads, masks, and staging; otherwise it may not reproduce the reported performance gain.

Would you be open to keeping the lowering as an internal backend mechanism, removing the Python-facing API, and adding a pattern-based rewrite with a safe fallback? This would preserve the optimization while keeping backend-specific fragment-layout details out of the user API.

zeroherolin:
@sunnycase Thanks — yes, I'm open to that: internal lowering, no Python-facing API, pattern-based rewrite with a fallback. I built the `join -> permute -> reshape` version to check it before agreeing, and it does regress badly as written. The cause turns out to be step 2 of your plan rather than the chain itself, so I'd like to propose one change there before I redo the PR.

## Measuring the chain

E=256, H=4096, I=256, topk=6, H20-3e, `do_bench_cudagraph` median:

| M | 8-segment (µs) | primitive (µs) | join chain (µs) | prim / base | chain / base |
|--:|--:|--:|--:|--:|--:|
| 16 | 126.98 | 118.49 | 271.78 | 1.072 | 0.467 |
| 136 | 310.46 | 289.43 | 716.07 | 1.073 | 0.434 |
| 256 | 379.90 | 350.12 | 875.97 | 1.085 | 0.434 |
| 512 | 489.38 | 459.79 | 551.84 | 1.064 | 0.887 |

Slower than the primitive, and slower than the unoptimised 8-segment loop it replaces. In the full kernel the chain ends at a `#linear` layout and gets staged:

```
%bs_full     = tt.reshape %t : ... -> tensor<64x128xbf16, #linear>
%bs_full_210 = ttg.local_alloc %bs_full
%bs_full_211 = ttg.local_load  %bs_full_210 : ... -> dot_op<kWidth=16>
```

256 `st.shared` per K-tile against 0 for the primitive. That round trip is the whole gap.

## On your question about layout propagation

You asked whether existing layout propagation can already eliminate the materialisation for some configurations, and whether the round trip is specific to enhanced RLC, MMAv2, or the full consumer IR. It can eliminate it, and the last of those three is the answer: on a small standalone kernel the `tt.reshape` emits `dot_op` directly and the two routes are indistinguishable. The materialisation is not intrinsic to the chain — only to when the compiler sees it.

Which is why I'd change step 2. Running the matcher after MMA layout selection is past the point where it can help: `Coalesce` and `AccelerateMatmul` have already committed the `dot_op`, blocked and `#shared` encodings, and rewriting the chain afterwards doesn't make them reconsider. Folding it **before `Coalesce`**, while the IR is still layout-free, should leave every later pass looking at a single op — structurally the IR the explicit primitive produces.

One more reason not to match after `AccelerateMatmul`: a join tree perturbs `kWidth` and the `#shared` swizzle through `computeOrigBitWidth` there, whose comment already flags why — it compensates for a single `JoinOp`, and merging 8 segments needs a 3-level tree. That isn't what costs the time here, but it does mean the pass has seen the chain and made decisions from it.

## On the fallback (your step 5)

Agreed that it's needed, and I think it has to be a second pass rather than "leave the IR unchanged on failure". Whether the concatenation can lower to a register relabel depends on the encoding layout propagation assigns — `dot_op`, `dim` being the contraction axis, `K >= 8*kWidth`, identical lane/warp/block bases, no broadcast along the indexed dimension. A matcher running before layout assignment can't check any of it, so "if any check fails, leave the original IR" isn't available to it: by the time the checks are decidable, the rewrite has already happened.

The failure mode matters because the op would be compiler-inserted. If the lowering rejects what it can't handle, a legal kernel — one that merely stores the concatenated result instead of feeding a dot — stops compiling, and the user never wrote the op. So: match early, and a second pass after the layouts settle that rebuilds the join tree wherever the relabel doesn't apply. Slower than the relabel in those cases, which is the correct trade.

## On bijectivity (your step 3)

This one can't be required literally. `dot_op` broadcasts along K over lanes and warps whenever the A operand is shared across warps, so the register-to-element map isn't injective. Sweeping `(M, warpsPerCTA)` with `gluon.to_linear_layout`, only `M=128, [4,1]` and `M=8, [1,1]` come out injective; `M=16, [4,1]` — the main decode shape here — has two zero warp bases, as do `M=8` with `[4,1]`, `[1,4]` and `[2,2]`. A literal check rejects the case the optimisation exists for.

The two conditions in this PR give the same guarantee and stay decidable: on the destination side, every result register slot must resolve to some source slot; on the source side, the fragment layout must not broadcast along the indexed dimension. The second isn't cosmetic — without it, `M ≤ 8` with non-`[1,1]` `warpsPerCTA` passes everything else while coordinate collisions silently duplicate source slots.

## On not auto-merging the eight dots

Agreed, and neither route does. The single `tl.dot` and the single activation load are written by the author:

```python
# before: one tl.dot per K sub-tile, weight operand fixed by the packing
for j in tl.static_range(8):
    k_off = k_logical_base + j * BLOCK_SIZE_K_PACK
    a_j_ptrs = a_base + (k_off + offs_ak_pack[:, None]) * stride_ak
    a_j = tl.load(a_j_ptrs, mask=token_mask[None, :], other=0.0)
    accumulator = tl.dot(bs[j], a_j, acc=accumulator)

# after: one activation load, one mma; only the concat spelling differs between routes
bs_full = tle.concat_dot_fragments(bs, dim=1)
a_full_ptrs = a_base + (k_logical_base + tl.arange(0, BLOCK_SIZE_K)[:, None]) * stride_ak
a_full = tl.load(a_full_ptrs, mask=token_mask[None, :], other=0.0)
accumulator = tl.dot(bs_full, a_full, acc=accumulator)
```

So there's no accumulation legality to prove and no segmented loads, masks or staging for the compiler to merge — collapsing the loop is the author's work under either route. The matcher only takes over how the concatenation is spelled, which is also why the reported gain carries over: the restructuring that produces it is unchanged.

## On the constraint layering

Worth keeping whichever route we take. The verifier holds what's decidable at any point in the pipeline: fragment count, `dim` range, result rank, element type, encoding, shape. The layout-dependent conditions can't move there, since `dot_op` encodings only exist after layout propagation and checking earlier would reject legal programs mid-pipeline.

Conditions that only cost performance — canonical register order, `splitRSDot` not undone, residual `convert_layout`, register pressure — are deliberately left unasserted: the compiler has legal fallbacks, and asserting them turns "works but slower" into "doesn't compile".

## Plan

If the two changes above look right to you — match before `Coalesce` rather than after MMA layout selection, and the fallback as a second pass — I'll prepare it as a separate PR against `main` and close this one. The register-gather lowering carries over unchanged, so the review you've already done on it applies either way.

zeroherolin:
Refer to https://github.com/flagos-ai/FlagTree/pull/892

PR #892, opened 2026-08-03
[Nvidia][TritonGPU] Fold an ordered K concatenation into one dot operand

zeroherolin:
Supersedes #881. Same lowering, no new Python API — the concatenation is now recognised in the backend from ops that already exist, per the review there.

## Problem

A dequant that unpacks several weights per `int32` produces sub-tiles that are adjacent along K but separate values. `tl.dot` takes one tensor per operand, so a kernel has to issue one mma per sub-tile, repeating the address arithmetic, the masked activation load and the operand staging once per segment. That overhead is fixed per iteration and does not shrink with the token count, so it dominates exactly where MoE decode shapes live.

Writing the ordered concatenation with existing ops does not fix this on its own. A balanced `tt.join` tree followed by a transpose and a reshape ends at a layout that distributes elements over lanes differently from the one the mma wants, so the operand is staged through shared memory and the loop ends up *slower* than the segmented version it replaces.

## Approach

Add `ttg.concat_dot_operand` and fold the chain into it **before any layout is assigned**. For a `dot_op` layout the wider fragment only adds register bases along K and leaves the lane, warp and block bases untouched, so every result element already sits on the thread holding its source, and the op lowers to a per-thread register relabel.

Folding early is what makes this work. Once `Coalesce` and `AccelerateMatmul` have committed the `dot_op`, blocked and shared encodings, rewriting the chain no longer changes what they picked — see *Why the match runs before layout assignment* below for the measurement.

Whether the relabel applies follows from the encoding layout propagation assigns, which the matcher cannot see. So a second pass runs once the layouts are final and rebuilds the join tree wherever it does not. Without it, a kernel that merely stores the concatenated result instead of feeding a dot would stop compiling: the op is compiler-inserted, so rejecting it in the lowering means refusing a program that was legal before the rewrite.

Three pieces:

- `ttg.concat_dot_operand`, an internal op with no Python binding
- `tritongpu-concat-dot-operand`, folding join tree → transpose → reshape before `Coalesce`
- `tritongpu-expand-concat-dot-operand`, restoring the chain after the final canonicalizer when the layouts do not permit the relabel

Both passes and the lowering share one predicate, `getConcatDotOperandRegisterMap`, which returns the result-register to fragment-register map, so they cannot disagree about what is lowerable.

## Why the match runs before layout assignment

This is the one substantive departure from the review suggestion in #881, which proposed matching after the MMA layout is selected. Measured with the matcher placed after `AccelerateMatmul`, the join/transpose/reshape chain is slower than the unoptimised 8-segment baseline it replaces:

| M | 8-segment baseline (µs) | matched after AccelerateMatmul (µs) | ratio |
|--:|--:|--:|--:|
| 1 | 22.45 | 33.38 | 0.67 |
| 16 | 127.13 | 271.81 | 0.47 |
| 136 | 310.53 | 716.33 | 0.43 |
| 256 | 381.51 | 877.37 | 0.44 |
| 512 | 493.04 | 552.90 | 0.89 |

The cause is `computeOrigBitWidth` in `AccelerateMatmul.cpp`, which halves `origBitWidth` once when it finds a `JoinOp` in the backward slice — its own comment notes this "won't be optimal if there is a tree of multiple JoinOps". A three-level tree is seven joins and the compensation divides by two once. That value feeds `minBitwidth` into `convertDotOperandForMMA`, which sets both the `dot_op` kWidth and the shared-memory swizzle, so the global→shared copy loses width: 4B `cp.async.ca` goes 12 → 60, 16B `cp.async.cg` goes 18 → 6. Only the MMAv2 branch reaches that path, which is why M=512 (wgmma) drops to 0.89× while M ≤ 256 drops to 0.43×.

Matching before `Coalesce` sidesteps this: every later pass sees a single op, and the IR is isomorphic to what an explicit primitive would have produced.

## Constraints

`getConcatDotOperandRegisterMap` derives the mapping on lane 0 and applies it to every thread, so it returns failure unless all of the following hold:

| Check | Why |
|---|---|
| encoding is `DotOperandEncodingAttr` | other encodings change lane ownership as `dim` grows |
| `dim` is the contraction axis (`opIdx=0` → last, `opIdx=1` → second-to-last) | extending M or N changes which lane owns an element |
| `kWidth == 0 \|\| K_frag >= 8 * kWidth` | below that the layout stops scaling with K, so the fragment is not a K-slice of the wider one |
| lane / warp / block bases identical | lane 0's mapping cannot otherwise be reused |
| register count conserved | mismatched element counts |
| `fragIdx` in range | out-of-bounds coordinate |
| every result slot inverts to a source slot | the encoding is not a clean K superset |
| fragment layout does not broadcast along the indexed dim | broadcasting maps several registers to one element, dropping some and duplicating others |

The first three cannot live in the op verifier: `dot_op` is only assigned after `AccelerateMatmul`, so before that point the tensors are necessarily `#blocked`, and verifying there would reject legal programs mid-pipeline.

The bijectivity check suggested in #881 would be too strong. `broadcastedDotOperandLayout` uses `zeros1D` for lane/warp along K so the A operand can be shared across warps, so the layout is often not bijective: scanning `(M, warpsPerCTA)` with `gluon.to_linear_layout`, `M=16, warps=4x1` has two zero warp bases, and `M=8` is non-bijective for `4x1`/`1x4`/`2x2`/`8x1`. Only `M=128, warps=4x1` and `M=8, warps=1x1` happen to be. The last two checks above are what bijectivity was standing in for, and they admit the decode shapes that matter.

## Isolation

No other backend schedules either pass, so `ttg.concat_dot_operand` never appears in their IR. Verified against the parent commit built separately, on H20-3e:

- **IR, byte for byte.** Eight unrelated kernels (vector add, softmax, matmul in f16/bf16/f32, a flash-attention shape, a join+split round-trip, layernorm) are identical at TTGIR, LLIR and PTX.
- **lit.** No change in the existing suite; the added `concat-dot-operand.mlir` passes.
- **Timing.** `mm` 2048³ 145.96 → 146.01 µs, softmax 4096² 32.02 → 32.02 µs, rms_norm 4096² 20.84 → 20.82 µs.

One property of the matcher is worth stating since it is not visible from the op: `onlyFeedsDot` follows `ConvertLayoutOp` and stops at anything else, so a dtype cast between the reshape and the dot makes the match fail. Fragments must already carry the dot's element type when they enter the join tree. Failing to match is safe — the chain compiles unchanged and stays correct — it just does not get the optimisation.

## Tests

- `test/TritonGPU/concat-dot-operand.mlir` — matching, the fold, and a fold-then-expand round-trip that must restore the same chain
- `test/TritonGPU/invalid.mlir` — verifier rejections: fewer than two fragments, wrong result shape, `dim` out of range
- Negative matching cases: a non-concat transpose order, an `allow_reorder` reshape, a chain whose result does not reach a dot, a join used elsewhere

On GPU: the cases that must *not* fold (join → store, wrong permute order, an escaping use) produce correct results with no spurious fold, and the expand fallback is exercised at K/frag=8 with 1, 4 and 8 warps — it rebuilds the join tree, compiles and computes correctly, while K/frag ≥ 16 folds. The join/dot/reshape/permute subset of `test_core.py` passes, 2592 tests.

## Results

Measured on a fused MoE GEMM over 53 deduped DeepSeek-V4-Flash trace shapes (E=256, H=4096, I=256, top_k=6) on H20-3e, bf16, `do_bench_cudagraph` median. mma count per K tile goes 8 → 1 in the plain kernels and 16 → 2 in the fused silu ones. Arithmetic mean speedup against vLLM Marlin 1.062 → 1.171, call-weighted 1.110 → 1.210, call-weighted total latency 8181.1 ms → 7792.9 ms (−4.74%). Every shape improves and output is bit-identical to the segmented path.

The consumer side is FlagGems flagos-ai/FlagGems#5140.


test/TritonGPU/concat-dot-operand.mlir
zhzhcookie, line 1: Please add copyright.

zeroherolin: Done.

zhzhcookie (approved):
LGTM

PR #924, opened 2026-08-06
fix(packaging): use system nlohmann-json

shiptux:
## Summary

- install distro nlohmann-json headers in Debian and RPM wheel-builder stages
- pass `JSON_SYSPATH=/usr` through the existing FlagTree dependency override so setup does not download nlohmann-json
- retain pinned downloads for backend-specific LLVM and device tools

## Verification

- confirmed `JSON_SYSPATH=/usr` resolves to CMake `JSON_INCLUDE_DIR=/usr/include` in the existing setup logic
- verified Fedora/Rocky use `json-devel`, while openEuler uses `nlohmann-json-devel`; the RPM builder handles both
- `git diff --check` passes

The full FlagTree wheel build takes 30-50 minutes and is left to repository CI.


zhzhcookie (approved):
LGTM

PR #930, opened 2026-08-06
[CI][BENCHMARK] Add mtheads3.6 qwen benchmark to ci

zhzhcookie:
Add mtheads3.6 qwen benchmark to ci, and update nv3.6 qwen benchmark path.


Galaxy1458 (approved):
LGTM

PR #944, opened 2026-08-10
[TLE][MTHREADS] Support barrier/wgmma/pipe/ws primitives

Kylin1207:
### 1.1 `buffer.slot(index)` and `barriers[index]`

`buffer.slot(index)` creates a view of one element of the leading stage
dimension. Constant and scalar dynamic indices are supported. Constant indices
are bounds checked during compilation. A block tensor is not accepted as a
stage index. Dynamic indices are not runtime bounds checked by this API, so the
kernel must keep them in range.

Barrier indexing follows the same static-or-scalar-dynamic model. Dynamic
indexing is kept as integer hardware-ID arithmetic; it does not create a
shared-memory barrier table.

### 2.1 Explicit TME completion with barrier

The supported explicit form is:

```python
full = tle.gpu.alloc_barriers(
    STAGES,
    arrive_count=1,
    init=tle.gpu.PENDING,
    expect_bytes=BLOCK_M * BLOCK_K * ELEMENT_BYTES,
)

tle.gpu.copy(
    a_desc,
    a_smem.slot(slot),
    (BLOCK_M, BLOCK_K),
    (m_offset, k_offset),
    barrier=full[slot],
)
tle.gpu.barrier_wait(full[slot], phaseIdx=phase)
```

The lowering emits transaction-byte registration, the TME load, and completion
arrival against the same hardware barrier ID. In a non-warp-specialized kernel,
thread 0 issues this sequence. In the supported warp-specialized form, the
issue thread is the first thread of producer partition 0.

Restrictions:

- Only descriptor global-to-shared TME copy accepts `barrier=`.
- A barrier array with more than one element must be explicitly indexed.
- `expect_bytes` must be a positive compile-time integer.
- The byte count is the complete logical transaction size and is not inferred
  by the standalone `copy(..., barrier=...)` API.
- A pipe-managed TME copy must not also carry an explicit user barrier.
- The explicit-completion copy must be in producer partition 0 when it is
  nested under the supported static `warp_specialize` form.

For an SQMMA-compatible grouped TME load whose leading width exceeds 256 bytes,
the current late lowering may emit multiple TME load intrinsics. Those emitted
segments receive the same logical `bar_id`. This does not relax the pipe
restriction that one acquire/commit interval contains exactly one logical TME
copy operation.

### 3.1 Allocation and resource model

`tle.gpu.alloc_barrier` and `tle.gpu.alloc_barriers` lower to ranges of MUSA
hardware barrier IDs. The barrier objects themselves do not consume shared
memory.

Hardware IDs are reserved from the mthreads function-wide resource pool. The
valid allocatable range is IDs 1 through 63. Explicit user barriers, pipe full
and empty rings, and warp-partition synchronization all consume this shared
budget. Compilation fails with a stable diagnostic when the combined demand
cannot be reserved.

All barrier allocations must be in the function entry block. When static warp
specialization is present, they must appear before the top-level
`warp_specialize`. All allocations must dominate the first barrier use. The
lowering initializes every allocated hardware barrier and emits one local CTA
rendezvous after initialization so that no partition can use an uninitialized
ID.

### 3.2 Phase-aware wait and arrive

The mthreads backend requires `phaseIdx`:

```python
tle.gpu.barrier_wait(barriers[slot], phaseIdx=iteration_phase)
tle.gpu.barrier_arrive(barriers[slot], arrive_count=1,
                       phaseIdx=iteration_phase)
```

The logical phase is reduced to one bit. The effective hardware phase is:

```text
(phaseIdx & 1) XOR initial_polarity
```

where `initial_polarity` is zero for `PENDING` and one for `READY`. Therefore,
`PENDING` and `READY` define the logical interpretation of generation zero;
they are not separate software state objects. Reusing a slot requires the
caller or pipe iteration mapping to advance/toggle the phase.

Static and scalar dynamic phases are supported. A block phase or non-integer
phase is rejected.

### 3.3 Arrival-count restriction

The allocation-time `arrive_count` is the number of arrivals required to
complete a hardware barrier generation and may be greater than one. However,
each emitted `barrier_arrive` operation can contribute only one arrival:

```python
tle.gpu.barrier_arrive(bar, arrive_count=1, phaseIdx=phase)
```

Passing an operation-level `arrive_count` other than one is rejected because
the current mthreads hardware lowering has no increment-by-N arrival operation.
For a consumer group, completion is achieved by convergent per-warp arrivals
against a barrier initialized with the consumer warp count.

### 3.4 Unsupported named-barrier path

The public frontend interprets `barrier_wait(bar)` or `barrier_arrive(bar)`
without `phaseIdx` as the generic named-barrier path. mthreads hardware barriers
are phase-aware resources and the backend does not implement the NVIDIA-style
named-barrier mechanism. Consequently, the no-`phaseIdx` form is rejected on
mthreads with the diagnostic that the named barrier backend is unsupported.

## 4. `tle.gpu.wgmma`  and  `tle.gpu.wgmma_wait`

The public API keeps the generic TLE spelling, but the mthreads frontend emits
backend-local operations:

```text
tle.gpu.wgmma      -> musa_tle.sqmma      -> ttmg.squad_dot
tle.gpu.wgmma_wait -> musa_tle.sqmma_wait -> ttmg.squad_dot_wait
```

The final LLVM IR contains MUSA SQMMA intrinsics and the hardware parameterless
SQMMA wait.

The currently supported contract is:

- A and B are rank-2 shared-memory `buffered_tensor` values.
- A and B have the same input dtype.
- Input dtype is `tl.float16`, `tl.bfloat16`, or `tl.float8e4nv`.
- The accumulator and output dtype are `tl.float32`.
- `max_num_imprecise_acc` is zero.
- `trans_a=False` and `trans_b=False`.
- M and N are at least 16 and divisible by 16.
- K is at least 16 and the complete M/N/K shape must admit a native PH1 SQMMA
  configuration.
- The active SQMMA warp count is at least four and divisible by four.
- Root A/B allocations use `layout=None, nv_mma_shared_layout=True` so the
  lowering can select a TME-compatible physical swizzle.

The compiler selects a supported native instruction shape and warp
distribution. A larger logical operation can be decomposed automatically. For
example, a logical `256 x 256 x 64` FP16 operation with 16 warps is covered by
a runtime test while the source contains one `wgmma` and one `wgmma_wait`.

An async result may feed the accumulator of another SQMMA or travel through the
supported loop-carried path, but it must pass through `wgmma_wait` before an
ordinary tensor consumer or store.

### 4.1 Wait limitation

Only the fully draining form is supported:

```python
```

`pendings > 0` is rejected. The initial implementation cannot keep multiple
independently addressable outstanding SQMMA groups and wait until only a
nonzero number remain. The returned accumulator is the dependency-released SSA
value and must be used after the wait.

### 4.2 Covered SQMMA types and shapes

Runtime coverage includes:

- FP16 with K in `{16, 32, 64}`;
- BF16 with K in `{16, 32, 64}`;
- FP8 E4M3 (`tl.float8e4nv`) with K in `{32, 64, 128}`;
- M/N pairs `(16,64)`, `(32,32)`, `(32,64)`, `(32,128)`, `(64,16)`,
  `(64,32)`, `(64,64)`, `(64,128)`, `(128,32)`, `(128,64)`, and
  `(128,128)`;
- a loop-carried accumulation case;
- the `256 x 256 x 64` macro-tile decomposition case.

This test matrix is evidence for the listed configurations, not a guarantee
that every shape satisfying only the divisibility checks can be mapped. The
native configuration selector remains authoritative and rejects an unmappable
shape during compilation.

### 5.1 Supported API model

The initial mthreads pipe is a CTA-scoped, cyclic, single-producer/
single-consumer pipe with exactly one payload field:

```python
a_smem = tle.gpu.alloc(
    (STAGES, BLOCK_M, BLOCK_K),
    dtype=tl.float16,
    layout=None,
    nv_mma_shared_layout=True,
)

a_pipe = tle.pipe(
    capacity=STAGES,
    scope="cta",
    name="a",
    a=a_smem,
)

writer = a_pipe.writer()
reader = a_pipe.reader()

write_slot = writer.acquire(iteration)
tle.gpu.copy(a_desc, write_slot.a, (BLOCK_M, BLOCK_K), offsets)
writer.commit(iteration)

read_result = reader.wait(iteration)
tile = read_result.slot.a
# consume tile
reader.release(iteration)
```

`slot.a` is a view created from the current stage index; it is not an
independent allocation. The stage and phase are derived from the iteration:

```text
stage = iteration % capacity
phase = (iteration // capacity) % 2
```

The lowerer creates two hardware-barrier rings for each pipe:

- `full[capacity]`, initialized logically `PENDING`, with one producer/TME
  completion arrival;
- `empty[capacity]`, initialized logically `READY`, with an arrival count equal
  to the consumer warp count.

The lifecycle mapping is:

| Pipe operation | Hardware protocol |
| --- | --- |
| `writer.acquire(i)` | Wait for `empty[stage]` using the writer-side phase. |
| TME copy + `writer.commit(i)` | Attach the copy to `full[stage]`, register transaction bytes, and complete the barrier when TME finishes. |
| `reader.wait(i)` | Wait for `full[stage]`. |
| `reader.release(i)` | Each consumer warp contributes one arrival to `empty[stage]`. |

### 5.2 Pipe restrictions

The initial mthreads pipe intentionally rejects:

- more than one payload field in one pipe;
- named readers or multiple readers;
- reader field subsets;
- `one_shot=True`;
- `writer.close`;
- scopes other than `scope="cta"`;
- an explicit barrier on a pipe-managed copy;
- zero or more than one logical TME copy between a matching
  `writer.acquire` and `writer.commit`;
- transaction sizes that vary between commits on the same pipe;
- an acquire/commit or wait/release pair that cannot be matched in the same
  block and for the same stage;
- mixing non-warp-specialized endpoints and warp-specialized endpoints on the
  same pipe.

Because close is unsupported, `reader.wait(...).is_closed` is materialized as
constant false on the mthreads path and must not be used as an end-of-stream
signal.

The payload field must be shared memory, rank two or higher, and its leading
dimension must equal `capacity`. Capacity must be a positive compile-time
integer. Capacities 1, 2, and 3 have compile and runtime coverage. Larger
capacities are not categorically rejected, but they consume two hardware
barrier IDs per pipe stage and have not received the same runtime coverage.

Multiple independent single-field pipes are supported. The optimized MM uses
one pipe for A and one for B because each input has an independent TME
transaction. This preserves one logical TME transaction per full barrier and
avoids unsupported multi-field aggregation.

### 5.3 Pipe resource cost

The full and empty barrier rings use hardware IDs and add no shared-memory
bytes. The shared-memory cost is the payload allocation itself. For two input
pipes with stage capacity `S`, the barrier demand is `4 * S` IDs before adding
any other user or partition-synchronization barriers.

### 6.1 Supported static form

The supported model has one default consumer and one worker producer:

```python
tle.gpu.warp_specialize(
    (
        (consumer, consumer_args),
        (producer, producer_args),
    ),
    worker_num_warps=(PRODUCER_WARPS,),
    worker_num_regs=(PRODUCER_REGS,),
)
```

The kernel launch `num_warps` belongs to the default consumer partition.
`worker_num_warps[0]` belongs to the producer. The physical launch uses their
sum. Producer warp count is not hard-coded to four; runtime coverage includes
producer counts 1, 2, 4, and 8, including an equal 4-consumer/4-producer case.

The lowering preserves the common `ttg.warp_specialize` container until late
lowering, then emits two explicit thread predicates:

```text
if tid >= consumer_num_warps * 32:
    producer

if tid < consumer_num_warps * 32:
    consumer
```

These are two explicit conditional branches, not an `if/else` dispatch and not
a switch. The producer predicate appears first in CFG order and rejoins before
the consumer predicate; this CFG shape does not serialize producer and consumer
threads. Hardware barrier IDs used by partition synchronization are
rematerialized in the partition, so the control path does not require a
shared-memory mailbox.

### 6.2 Warp-specialize restrictions

The initial implementation requires:

- exactly one marked `warp_specialize` operation per module;
- exactly one default consumer and one producer partition;
- no values returned by the `warp_specialize` operation;
- placement in the Triton function entry block;
- placement as the final operation before `tt.return`;
- one block in each partition region;
- compile-time sequences for `worker_num_warps` and `worker_num_regs`;
- a positive producer warp count accepted by the common dialect verifier;
- captures that dominate the partition split;
- producer-side explicit-completion TME operations only in worker partition 0.

Multiple producers, multiple consumers, nested/static WS regions, partition
results, nonterminal WS placement, and dynamic warp counts are not supported.
The default and producer functions are emitted inline for mthreads, so this is
not an independent-device-kernel launch model.

### 6.3 Synchronization and shared-memory cost

The supported static WS control-flow lowering itself adds no shared-memory
allocation. Pipe barriers and post-SQMMA partition synchronization use hardware
barrier IDs. A local CTA rendezvous is retained for barrier initialization.
Consequently, the important resource constraint is the 63-ID hardware barrier
budget, in addition to the actual payload shared memory and register usage.

## 7. End-to-end matrix-multiplication status

The repository contains an optimized TLE warp-specialized MM and a common
correctness/performance harness at:

```text
third_party/mthreads/python/test/unit/tle/benchmark/test_mm.py
```

It compares:

- native Triton MM;
- optimized TLE-WS MM;
- `torch.matmul`.

The TLE-WS path combines automatic SQMMA shared layouts, two single-field
pipes, explicit TME completion, `wgmma`/`wgmma_wait`, and one static
consumer/producer split. Pipeline stages 1, 2, and 3 are accepted. The pytest
default correctness/performance case is `1024 x 1024 x 1024`; the command-line
harness also provides larger default shapes. Those larger presets are
benchmark inputs, not a claim that every preset is run in routine unit tests.

Example invocations:

```bash
python -m pytest -q -s \
  third_party/mthreads/python/test/unit/tle/benchmark/test_mm.py
```

```bash
python third_party/mthreads/python/test/unit/tle/benchmark/test_mm.py \
  --shape 1024 1024 1024 \
  --shape 8192 7168 16384 \
  --stages 1 2 3
```

The benchmark verifies numerical accuracy before reporting median latency and
the Triton/Torch, TLE-WS/Torch, and TLE-WS/Triton speed ratios.

| M | N | K | Stages | Torch (ms) | Triton (ms) | TLE-WS (ms) | Triton vs Torch | TLE-WS vs Torch | TLE-WS vs Triton |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1024 | 1024 | 1024 | 1 | 0.029480 | 0.072020 | 0.047400 | 0.4093x | 0.6219x | 1.5194x |
| 1024 | 1024 | 1024 | 2 | 0.016360 | 0.062120 | 0.030240 | 0.2634x | 0.5410x | 2.0542x |
| 1024 | 1024 | 1024 | 3 | 0.016480 | 0.055280 | 0.029020 | 0.2981x | 0.5679x | 1.9049x |
| 2048 | 2048 | 2048 | 1 | 0.056160 | 0.230500 | 0.159200 | 0.2436x | 0.3528x | 1.4479x |
| 2048 | 2048 | 2048 | 2 | 0.056520 | 0.188720 | 0.094000 | 0.2995x | 0.6013x | 2.0077x |
| 2048 | 2048 | 2048 | 3 | 0.055660 | 0.161240 | 0.086560 | 0.3452x | 0.6430x | 1.8628x |
| 4096 | 4096 | 4096 | 1 | 0.414000 | 1.142960 | 0.771680 | 0.3622x | 0.5365x | 1.4811x |
| 4096 | 4096 | 4096 | 2 | 0.416640 | 0.898240 | 0.462880 | 0.4638x | 0.9001x | 1.9405x |
| 4096 | 4096 | 4096 | 3 | 0.417000 | 0.793680 | 0.438000 | 0.5254x | 0.9521x | 1.8121x |
| 8192 | 8192 | 8192 | 1 | 3.354360 | 8.501160 | 5.928640 | 0.3946x | 0.5658x | 1.4339x |
| 8192 | 8192 | 8192 | 2 | 3.334400 | 6.929920 | 3.548240 | 0.4812x | 0.9397x | 1.9531x |
| 8192 | 8192 | 8192 | 3 | 3.346360 | 6.090460 | 3.507880 | 0.5494x | 0.9540x | 1.7362x |
| 8192 | 7168 | 16384 | 1 | 5.821160 | 13.915180 | 10.253881 | 0.4183x | 0.5677x | 1.3571x |
| 8192 | 7168 | 16384 | 2 | 5.835540 | 11.634160 | 6.170440 | 0.5016x | 0.9457x | 1.8855x |
| 8192 | 7168 | 16384 | 3 | 5.819780 | 10.192900 | 5.830360 | 0.5710x | 0.9982x | 1.7482x |

### DIFF
--- a/python/triton/experimental/tle/language/gpu/core.py
+++ b/python/triton/experimental/tle/language/gpu/core.py
@@ -306,6 +306,9 @@ def alloc(
         dtype: Data type
         layout: Memory layout encoding (optional)
         scope: Storage type (default to shared memory)
+        nv_mma_shared_layout: Select an MMA-consumer-defined shared layout when
+            ``layout`` is None. On mthreads this is materialized by the SQMMA
+            lowering rather than as an NVIDIA encoding.
         _semantic: Semantic analyzer (internal use)
 
     Returns:
@@ -330,7 +333,6 @@ def alloc(
         raise ValueError(f"Storage type must be tle.scope, but got {type(scope)}")
 
     layout = tl._unwrap_if_constexpr(layout)
-    auto_shared_layout = layout is None
     if layout is not None and not isinstance(layout, tle.shared_layout):
         # Handle constexpr None
         if hasattr(layout, 'value') and layout.value is None:
@@ -350,6 +352,8 @@ def alloc(
 
     # Map scope to storage (backward compatibility)
     storage = scope
+    mthreads_auto_sqmma_shared_layout = (mthreads_common.enabled() and storage == tle.smem
+                                         and mthreads_wgmma.use_auto_shared_layout(layout, nv_mma_shared_layout))
 
     try:
         unwrapped_shape = [tl._unwrap_if_constexpr(dim) for dim in shape]
@@ -363,7 +367,7 @@ def alloc(
 
         if layout is None:
             if storage == tle.smem:
-                if not nv_mma_shared_layout:
+                if mthreads_auto_sqmma_shared_layout or not nv_mma_shared_layout:
                     layout = tle.swizzled_shared_layout.make_default(rank=len(shape))
                     layout_handle = _semantic.builder.make_swizzled_shared_encoding_attr(
                         layout.vectorSize,
@@ -406,7 +410,7 @@ def alloc(
                 tensor_handle = _semantic.builder.create_local_alloc(mutable_ty, init_value.handle)
             else:
                 tensor_handle = _semantic.builder.create_local_alloc(full_shape, elem_type, layout_handle)
-            if auto_shared_layout and mthreads_common.enabled():
+            if mthreads_auto_sqmma_shared_layout:
                 mthreads_wgmma.mark_auto_shared_layout(_semantic.builder, tensor_handle)
         else:
             raise ValueError(f"Storage type {storage} not yet supported")
--- a/python/triton/experimental/tle/language/gpu/core.py
+++ b/python/triton/experimental/tle/language/gpu/core.py
@@ -27,6 +27,7 @@ from . import types as tle
 from .mthreads import common as mthreads_common
 from .mthreads import copy as mthreads_copy
 from .mthreads import warp_specialize as mthreads_warp_specialize
+from .mthreads import wgmma as mthreads_wgmma
 from .iluvatar import copy as iluvatar_copy
 from triton.compiler.code_generator import flatten_values_to_ir, unflatten_ir_values
 
@@ -328,6 +329,7 @@ def alloc(
         raise ValueError(f"Storage type must be tle.scope, but got {type(scope)}")
 
     layout = tl._unwrap_if_constexpr(layout)
+    auto_shared_layout = layout is None
     if layout is not None and not isinstance(layout, tle.shared_layout):
         # Handle constexpr None
         if hasattr(layout, 'value') and layout.value is None:
@@ -399,6 +401,8 @@ def alloc(
                 tensor_handle = _semantic.builder.create_local_alloc(mutable_ty, init_value.handle)
             else:
                 tensor_handle = _semantic.builder.create_local_alloc(full_shape, elem_type, layout_handle)
+            if auto_shared_layout and mthreads_common.enabled():
+                mthreads_wgmma.mark_auto_shared_layout(_semantic.builder, tensor_handle)
         else:
             raise ValueError(f"Storage type {storage} not yet supported")
 
@@ -789,16 +793,23 @@ def wgmma(
     """
     trans_a = _require_wgmma_bool(trans_a, "trans_a")
     trans_b = _require_wgmma_bool(trans_b, "trans_b")
-    a, b = _canonicalize_wgmma_operands(a, b, trans_a, trans_b, _semantic)
+    mthreads_enabled = mthreads_common.enabled()
+    if mthreads_enabled:
+        mthreads_wgmma.validate_operands(a, b, acc, trans_a, trans_b)
+    else:
+        a, b = _canonicalize_wgmma_operands(a, b, trans_a, trans_b, _semantic)
 
     m, k = [int(tl._unwrap_if_constexpr(dim)) for dim in a.type.shape]
     k_b, n = [int(tl._unwrap_if_constexpr(dim)) for dim in b.type.shape]
     if k != k_b:
         raise ValueError(f"wgmma shape mismatch: a is {a.type.shape}, b is {b.type.shape}")
-    if m < 64 or m % 64 != 0:
-        raise ValueError("wgmma result M dimension must be divisible by 64")
-    if n < 8 or n % 8 != 0:
-        raise ValueError("wgmma result N dimension must be divisible by 8")
+    if mthreads_enabled:
+        mthreads_wgmma.validate_dimensions(m, n)
+    else:
+        if m < 64 or m % 64 != 0:
+            raise ValueError("wgmma result M dimension must be divisible by 64")
+        if n < 8 or n % 8 != 0:
+            raise ValueError("wgmma result N dimension must be divisible by 8")
     if k < 16:
         raise ValueError("wgmma K dimension must be at least 16")
 
@@ -828,6 +839,9 @@ def wgmma(
         max_num_imprecise_acc = _require_wgmma_int(max_num_imprecise_acc, "max_num_imprecise_acc")
         if max_num_imprecise_acc < 0:
             raise ValueError("max_num_imprecise_acc must be non-negative")
+
+    if mthreads_enabled:
+        mthreads_wgmma.validate_options(max_num_imprecise_acc, out_dtype)
 
     ret_scalar_ty = _wgmma_ret_scalar_ty(a.dtype, out_dtype)
     ret_ty = tl.block_type(ret_scalar_ty, [m, n])
@@ -867,6 +881,8 @@ def wgmma_wait(pendings, acc=None, _semantic=None, _generator=None) -> tl.tensor
     pendings = _require_wgmma_int(pendings, "pendings")
     if pendings < 0:
         raise ValueError("wgmma_wait pendings must be non-negative")
+    if mthreads_common.enabled():
+        mthreads_wgmma.validate_wait_pendings(pendings)
     if not isinstance(acc, tl.tensor):
         raise ValueError(f"wgmma_wait acc must be a tl.tensor, got {type(acc).__name__}")
     result = _semantic.builder.create_tle_wgmma_wait(acc.handle, pendings)
--- a/python/triton/experimental/tle/language/gpu/core.py
+++ b/python/triton/experimental/tle/language/gpu/core.py
@@ -1083,9 +1083,12 @@ def copy(
             raise ValueError("copy barrier is only supported for TMA global-to-shared copy")
         return normcopy(src, dst, shape, direction, _semantic)
     if mthreads_enabled:
+        barrier_slot = None
         if barrier is not None:
-            raise ValueError("TMA copy barrier is only supported on NVIDIA backend")
-        return mthreads_copy.tmacopy(src, dst, direction, shape, offsets, _semantic)
+            if direction != CopyDirection.GM_TO_LOCAL:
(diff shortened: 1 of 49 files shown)


sunnycase (approved):
LGTM

zhzhcookie (approved):
LGTM

PR #956, opened 2026-08-11
[TLE] Fix SparseMLA local-pointer encoding propagation

sunnycase:
## Background

The NVIDIA CI SparseMLA tutorial started failing all five `tle-flashmla-prefill` benchmark cases after TLE helper inlining was enabled. The compiler stopped in `TritonTleSelectEncodings` with:

```text
'tt.load' op failed to verify that result matches ptr type
```

The benchmark then caught the compilation exception, returned NaN timings, and exited with status 0. As a result, GitHub Actions continued to the next test and the delayed/buffered benchmark output made the failure look like a hang.

Failing run: https://github.com/flagos-ai/FlagTree/actions/runs/30510009926/job/90767823379

## Root cause

Full-view loads from `tle.local_pointers` are intentionally excluded from consumer encoding votes because they are rewritten to `ttg.local_load` later in the pipeline. However, `TleSelectEncodings` also skipped those loads while propagating a newly selected pointer encoding to users.

When another user selected a different encoding for the local pointer, the pointer result type changed while the full-view `tt.load` result retained its old encoding. Verification therefore failed before `TleOptimizeLocalPointerLoads` had a chance to rewrite the load.

Separately, the SparseMLA table and single-benchmark paths treated compile/runtime exceptions like unsupported optional providers, printed an error, and continued.

## Solution

- Keep rewritable full-view loads out of encoding inference, but always synchronize their result type with the selected pointer encoding.
- Insert a `convert_layout` bridge so existing consumers retain their original result layout.
- Add a `-verify-each` MLIR regression covering a full-view load whose local pointer is retagged by another consumer.
- Propagate import, compilation, and runtime exceptions from forward/decode table and single benchmarks.
- Preserve skip/NaN behavior only when an optional provider is genuinely unavailable.
- Reject non-finite measured timings instead of silently reporting them.

This behavior is identical locally and in CI; no CI-only flag is required.

## Validation

- The new `test_tle_select_encodings.mlir` case passes with `-triton-tle-select-encodings -verify-each`.
- On H800, the minimal SparseMLA reproducer (`B=1, S=1, SKV=2048, topk=2048`) compiles, launches, and matches the Triton output.
- A synthetic compilation failure in the default table benchmark exits the Python process with status 1.
- Python syntax and diff checks pass.

The full TLE lit invocation reports 63/70 passing; the remaining failures are in unrelated alignment, remote-pointer, and WGMMA checks, while the added regression passes independently.

### DIFF
--- a/python/test/regression/test_sparse_mla_autotune_configs.py
+++ b/python/test/regression/test_sparse_mla_autotune_configs.py
@@ -1,6 +1,9 @@
 import inspect
 import importlib.util
+import math
 from pathlib import Path
+
+import pytest
 
 
 def _load_sparse_mla_module():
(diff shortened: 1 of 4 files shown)


zhzhcookie (approved):
LGTM

Galaxy1458 (approved):
LGTM

PR #939, opened 2026-08-08
[TLE][AMD] Add AMD tile and cumsum lowering

zihaomu:
## What

- Add AMD LLVM lowering for `tle.extract_tile`, `tle.insert_tile`, and `tle.exclusive_cumsum`.
- Register AMD shared-memory scratch allocation for the three operations.
- Route shared tile lowering through the existing `!targetInfo.isCuda()` boundary, preserving the CUDA/NVVM path without adding a target capability API.
- Use FlagTree backend identity helpers in the shared cumsum tests; HIP backends use the active target warp size while non-HIP backends retain the existing 32-thread test assumption.
- Reject unsupported TLE and accidental NVVM operations during AMD LLVM conversion.
- Fix clean parallel-build dependencies required by AMD targets.
- Add focused AMD correctness and `triton-opt + FileCheck` coverage.

## Why

The shared TLE lowering previously selected target-provided thread-id, barrier, and shuffle operations only for HCU. AMD implements those primitives but could not reuse the shared tile/cumsum lowering.

This PR is intentionally limited to AMD-owned implementation, the minimal shared predicates needed to select the non-CUDA path, and focused tests. It does not modify other backend-owned files.

## Testing

- Fresh `nvidia;amd`, `TRITON_BUILD_UT=ON`, `-j32` build: 379 steps completed.
- `TestOptimizeLDS`: 1/1 passed.
- Latest-head full incremental Ninja build and `libtriton.so` link: passed.
- gfx1201 cumsum suite: `9 passed, 4 expected skips` (NVIDIA PTX guard, HCU ISA guard, and two AMD local-pointer follow-ups).
- Backend policy checks: HIP wave32/wave64 uses target warp size; non-HIP warp sizes 12/128 retain 32 for these shared tests.
- AMD `triton-opt + FileCheck`: passed for extract, insert, and exclusive cumsum.
- Unsupported `tle.get_device_id` and accidental `nvvm.barrier0`: explicitly rejected during AMD conversion.
- Ruff 0.9.1, YAPF 0.43.0, clang-format 19.1.6, Python syntax, and `git diff --check`: passed.
- Rebased onto upstream `553537f6d`: conflict-free and 0 commits behind.

PR #972, opened 2026-08-12
[TLE][AMD] Add local pointers lowering and fix buffer ops address space

WhatGhost:
## What

- Register `populateLocalPointersOpToLLVMPatterns` in the AMD TLE-to-LLVM partial conversion so `tle.local_pointers` is legalized on AMD.
- Wire the five TTGIR local-pointer passes into the HIP pipeline right after `add_optimize_thread_locality`, matching the placement and order used by the NVIDIA and PPU backends.
- Reject pointers outside the global address space in `canUseBufferOps`, so shared-memory pointers are never promoted to AMD buffer ops.
- Add `triton-opt + FileCheck` coverage for both local-pointer forms, and for the buffer-ops address-space guard.
- Un-skip the two cumsum cases that #939 left behind as AMD local-pointer follow-ups.

## Why

#939 landed AMD lowering for `extract_tile`, `insert_tile`, and `exclusive_cumsum`, but `tle.local_pointers` remained unimplemented, so any kernel that stages data through `tle.gpu.local_ptr` failed with:

```
error: failed to legalize operation 'tle.local_pointers' that was explicitly marked illegal
```

`local_pointers` underpins the rest of the TLE-Struct GPU layer, so this blocked the GEMM, local-store and pipeline integration tests as well as the two cumsum cases that #939 marked as follow-ups.

**Registering the pattern alone is not sufficient.** `LocalPointersOpConversion` hard-`cast`s the memdesc encoding to `ttg::SharedEncodingTrait` and requires an encoding on the result tensor; that encoding comes from `add_select_encodings`, and the barriers guarding shared-memory traffic come from `add_insert_local_pointer_barriers`. The pass order is load-bearing and is kept identical to the other backends.

**Why the buffer-ops change is in scope.** With the pattern in place, lowering then failed inside `TritonAMDGPUConvertToBufferOps`:

```
error: 'amdg.buffer_store' op failed to verify that value element type matches the pointed type of ptr
```

Buffer ops address memory through a global-memory resource descriptor, but `canUseBufferOps` only validated the pointer *shape* (splatted base, 32-bit offset, 2GB range) and never the address space, so a shared-memory pointer passed the gate and produced an invalid op. The guard is added to `canUseBufferOps` itself because all four load/store/atomic/rmw patterns share it.

This defect is independent of TLE and reproduces from plain Triton IR with a `!tt.ptr<f32, 3>` kernel argument, which is why it is a separate commit. It is reached from TLE only by the scalar-index form of `local_ptr` (`local_ptr(buf, (0,))` followed by `splat` + `addptr`); the tensor-index form was already unaffected. Fixing it also repaired three pre-existing top-k failures unrelated to `local_pointers`.

Scope is limited to AMD-owned files. The only shared file touched is `python/test/tle/unit/test_tle_cumsum.py`, where the removed guard (`is_hip() and not FLAGTREE_BACKEND`) is constant-false for NVIDIA and HCU, so no other backend changes behavior.

## Testing

Environment: AMD Radeon Graphics (gfx1201, RDNA4, warp size 32), Ubuntu 24.04, `nvidia;amd` build with TLE enabled.

**Now passing (before → after):**

- TLE unit suite: 59 → 74 passed. All 13 previously failing cases (12 in `test_tle_gpu_local_ptr.py`, 1 in `test_tle_gpu_slot.py`) now pass.
- TLE integration suite: 1 → 10 passed (GEMM, local-store, 4 pipeline cases, 3 top-k).
- End-to-end: a `local_ptr` kernel now reaches `amdgcn` and produces correct results on device; previously it stopped at TTGIR.
- The two `test_tle_cumsum.py` cases marked `requires AMD local-pointer lowering` now pass and are un-skipped.

**Regression:**

- `python/test/unit/language/test_core.py`: 10466 passed, 631 skipped, 0 failed. This is the key check, since the buffer-ops guard affects every AMD kernel, not just TLE ones.
- AMD lit (`test/TritonGPU/amd`, `test/Conversion/amd`, `test/Analysis/amd`): 73/75. The 2 failures (`async_ops_to_llvm_gfx1250.mlir`, `cluster_load.mlir`) were confirmed pre-existing by stashing this branch's changes and re-running against the same base.

**Cross-architecture:** the new local-pointer FileCheck passes on gfx90a, gfx942, gfx950 (warp 64) and gfx1100, gfx1201, gfx1250 (warp 32); the lowering checks address-space semantics and carries no warp-size or ISA-generation assumptions.

**Lint:** `pre-commit run --files` over all touched files — ruff 0.9.1, yapf 0.43.0, clang-format 19.1.6, mypy 1.15.0 and the hygiene hooks all pass.

### DIFF
--- a/python/test/tle/unit/test_tle_cumsum.py
+++ b/python/test/tle/unit/test_tle_cumsum.py
@@ -27,10 +27,6 @@ _nv_mma_shared_layout = tl.constexpr(False if _is_hcu_backend() else True)
 threads_per_warp = get_current_target().warp_size if is_hip() else 32
 
 
-def _is_amd_hip_backend():
-    return is_hip() and not FLAGTREE_BACKEND
-
-
 def _require_cuda():
     try:
         if _is_enflame_backend():
@@ -280,7 +276,6 @@ def test_tle_cumsum_amdgcn_fastpath_regression_guard():
         "Detected predicated ds_write: possible regression to generic path"
 
 
-@pytest.mark.skipif(_is_amd_hip_backend(), reason="requires AMD local-pointer lowering")
 def test_tle_cumsum_helper_preserves_adjacent_sentinel():
     block = 512
     num_warps = block // threads_per_warp
@@ -302,7 +297,6 @@ def test_tle_cumsum_helper_preserves_adjacent_sentinel():
     torch.testing.assert_close(sentinel, expected_sentinel)
 
 
-@pytest.mark.skipif(_is_amd_hip_backend(), reason="requires AMD local-pointer lowering")
 def test_tle_cumsum_scalar_base_addptr_alias_regression():
     block = 512
     num_warps = block // threads_per_warp
--- a/test/Conversion/amd/tle_local_pointers_to_llvm.mlir
+++ b/test/Conversion/amd/tle_local_pointers_to_llvm.mlir
@@ -1,0 +1,70 @@
+// Copyright 2025-     FlagOS Contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files
+// (the "Software"), to deal in the Software without restriction,
+// including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software,
+// and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// RUN: triton-opt %s -split-input-file --allocate-amdgpu-shared-memory --convert-triton-amdgpu-to-llvm=arch=gfx1201 --convert-builtin-func-to-llvm | FileCheck %s
+
+#blocked = #ttg.blocked<{sizePerThread = [1], threadsPerWarp = [32], warpsPerCTA = [1], order = [0]}>
+#shared = #ttg.swizzled_shared<{vec = 1, perPhase = 1, maxPhase = 1, order = [0]}>
+#smem = #ttg.shared_memory
+
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 1 : i32, ttg.target = "hip:gfx1201", "ttg.threads-per-warp" = 32 : i32} {
+  // CHECK-LABEL: llvm.func @local_pointers_tensor_index
+  // The lowered pointers must stay in the shared address space and be reached
+  // through plain GEPs, with no NVVM op leaking into the AMD path.
+  // CHECK: llvm.mlir.addressof @global_smem : !llvm.ptr<3>
+  // CHECK: llvm.getelementptr {{.*}}!llvm.ptr<3>
+  // CHECK: llvm.load {{.*}}!llvm.ptr<3>
+  // CHECK: llvm.store {{.*}}!llvm.ptr<3>
+  // CHECK-NOT: tle.local_pointers
+  // CHECK-NOT: nvvm.
+  tt.func public @local_pointers_tensor_index(%idx: tensor<32xi32, #blocked>) {
+    %buf = ttg.local_alloc : () -> !ttg.memdesc<32xf32, #shared, #smem, mutable>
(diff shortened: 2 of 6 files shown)


WhatGhost:
Hi, team @zhzhcookie @sunnycase @Galaxy1458 @i3wanna2   , could you please help to review this PR?  this PR add  local pointers lowering and fix buffer ops address space based on #939

sunnycase:
Thanks for the detailed implementation and testing! Have you observed any performance improvements from the newly added `tle.local_pointers` lowering on specific operators or workloads, such as GEMM, cumsum, or top-k? If so, could you please share the relevant benchmark results?

sunnycase:
One more question: does the current `tle.local_pointers` implementation support `tl.atomic_*` operations (for example, `tl.atomic_add`) on pointers returned by `local_ptr`? If not, is this support planned?

WhatGhost:
@sunnycase Thanks you for your review! for the second question,``` local_ptr``` return ```tl.pointer_type(elem, 3)```, for my opinion, i think it's can support, but i need to check it. I'll relay the results of both benchmark and atomic_add after i checked. Maybe today or next week (depends on whether we can find an available machine for testing). Thanks again

WhatGhost:
@sunnycase Hi, for the benchmark
One clarification first: this PR is a functional enablement rather than an optimization. Before it, `tle.local_pointers` failed to legalize on AMD, so there is no meaningful before/after performance pair — the "before" simply did not compile. What I can offer is a comparison of the TLE `local_ptr` path against equivalent non-TLE implementations.

I ran the upstream TLE tutorials that exercise `local_ptr` and already carry `do_bench` harnesses. All of them were unrunnable on AMD before this PR. Environment: AMD Radeon Graphics (gfx1201, RDNA4, warp size 32).

**`python/tutorials/tle/03-topk.py`** — radix select staged through `local_ptr`, vs a pure-Triton top-k and `torch.topk` (ms, lower is better):

| M | N | K | Triton-RadixSelect (TLE) | Triton-TopK | Torch-TopK |
|---|---|---|---|---|---|
| 64 | 128 | 8 | 0.00988 | **0.00780** | 0.01396 |
| 64 | 1024 | 32 | 0.01256 | **0.01122** | 0.02288 |
| 64 | 8192 | 128 | **0.03642** | 0.04222 | 0.14020 |
| 128 | 32768 | 256 | **0.15516** | 0.21642 | 0.17948 |

The TLE version wins on the larger shapes (~14% faster than pure Triton at N=8192, ~28% at N=32768, and ~16% faster than Torch there), and loses slightly on the small ones where the staging overhead is not amortized. Crossover is around N≈8192.

**`python/tutorials/tle/01-fft.py`** (ms, lower is better):

| N | Triton | TLE | Torch |
|---|---|---|---|
| 64 | 0.06200 | **0.06008** | 0.04768 |
| 128 | 0.06724 | **0.05300** | 0.06728 |
| 256 | 0.09940 | **0.08488** | 0.08576 |
| 512 | 0.18576 | **0.17084** | 0.15044 |
| 1024 | 0.38810 | **0.35084** | 0.35068 |

Here the TLE version is faster than the pure-Triton one at every size (3–21%), and matches Torch from N=256 upward.

**Two tutorials still cannot run**, but for a reason outside this PR's scope: `python/tutorials/tle/02-moe_align_block_size.py` and `python/tutorials/tle/04-cluster-gemm.py` both fail with

```
error: failed to legalize operation 'tle.distributed_barrier' that was explicitly marked illegal
```

That is a separate TLE-Distributed op with no AMD lowering yet — unrelated to `local_pointers`, and it fails cleanly rather than crashing.  Which we will be implement on the future

WhatGhost:
@sunnycase And for the atomic op
the short answer is that it works for some forms of `local_ptr` and crashes the compiler for others. Here is what I measured on gfx1201 (RDNA4, warp size 32).

**Current state**

| Form | `tl.atomic_*` |
|---|---|
| `local_ptr(buf)` (full view) | works, lowers to a real LDS atomic (`ds_add_f32`) |
| `local_ptr(buf, (0,))` (scalar index) | works (`atomic_add`, `atomic_cas`) |
| `local_ptr(buf, (idx,))` (tensor index) | **compiler segfault** (`atomic_add`, `atomic_max`) |

So it is partially supported today, and the tensor-index form fails hard rather than reporting an error.

**Why**

`AtomicRMWOpConversion` in `third_party/amd/lib/TritonAMDGPUToLLVM/LoadStoreOpToLLVM.cpp` dereferences the axis info and indexes it without checking:

```cpp
unsigned contigWithinLanes =
    axisAnalysisPass.getAxisInfo(ptr)->getContiguity(threadOrder.front());
```

`getAxisInfo` may return `nullptr`, and a default-constructed `AxisInfo` has an empty `contiguity` vector (rank 0), so indexing it is out of bounds. `tle.local_pointers` has no `AxisInfoVisitor`, so the analysis produces exactly such a degenerate entry and this line faults.

Two things worth noting:

- The safe accessor `ModuleAxisInfoAnalysis::getContiguity` already guards this case for TLE (`#ifdef __TLE__` in `lib/Analysis/AxisInfo.cpp`), but this particular call site bypasses it. `getVectorSize()` in the same function does go through the safe accessor, which is why it does not fault.
- `contigWithinLanes` only refines `enableIntraWaveReduce`, which is restricted to `{CDNA3, CDNA4}` and is therefore already `false` on RDNA. So on gfx1201 the crash happens while computing a value that is never used.


For the record it is also not a regression from this PR or #939 — the unguarded dereference comes from upstream `020cb91305` (triton-lang/triton #6258, "[AMD] NFC: refactor atomic rmw lowering patterns"). Upstream never hits it because every native pointer-producing op has an `AxisInfoVisitor`. This PR is just the first thing that makes the path reachable on AMD.

**Plan**

The fix is a small guard at that call site (fall back to leaving the intra-wave reduce optimization disabled when the axis info is degenerate, so CDNA behaviour is untouched). Since it is an AMD backend defect independent of TLE, I would rather not widen this PR — I will send it as a separate `[AMD]` PR after this is merge. is that OK?

zhzhcookie (approved):
LGTM

sunnycase (approved):
Thanks for your clarification! @WhatGhost 
LGTM

PR #978, opened 2026-08-13
[BACKEND][SUNRISE] update sunrise backend plugin.

xdali-adora:
[BACKEND][SUNRISE] update sunrise backend plugin.


zhzhcookie (approved):
LGTM

PR #988, opened 2026-08-14
[CI] Exclude vendor setup utils from CORE and map to vendor labels

liuhaoyutz:
- Add negated patterns to prevent vendor-specific setup files under python/setup_tools/utils/ from being incorrectly labeled as CORE.
- Add those same files to their respective vendor labels.


zhzhcookie (approved):
LGTM

PR #1001, opened 2026-08-17
[QC][TLE] Support cross-dtype Hopper WGMMA accumulator reuse

July-h5kf3:
## Problem

A TLE Hopper WGMMA chain can reuse the same FP32 accumulator for FP8/K32 and BF16/K16 operations. The accumulator retains its previous encoding, so selecting the instruction shape entirely from that encoding can incorrectly emit BF16/K32 and fail during PTX lowering.

The failure was reproduced on H800 with the PR parent compiler (`fd3ea294d`) using the unmodified downstream MLA operator at B1/L640/HQ128:

```text
NVGPUToLLVMPass.cpp:496:
Assertion `supported && "WGMMA type or shape is not supported"' failed.
RuntimeError: PassManager::run failed
```

The BF16 WGMMA consumes the preceding FP8 accumulator with `instrShape = [16, 64, 32]`; the old lowering incorrectly selects K32 instead of the BF16-required K16. With this fix (`2ace55eca`), the same operator source compiles and passes output/LSE accuracy checks using a separate fresh compilation cache.

## Changes

- Select instruction K from the current operand types for both register-A and shared-A operations, while preserving the accumulator's M/N and layout.
- Reuse the existing operand-type checks, including TF32 precision validation and support for mixed E4M3/E5M2 FP8 operands.
- Validate register-A layout compatibility independently of the chain marker, which FenceInsertion can remove. Incompatible layouts now produce a diagnostic even when the marker is absent.
- Keep the lowering changes scoped to TLE and simplify instruction-shape construction.

## Validation

On H800:

- Build and pre-commit checks passed.
- 14 MLIR scenarios passed, covering cross-dtype reuse, supported operand types, and rejection of incompatible layouts/types, including cases without the chain marker.
- 6 routing tests and 3 downstream operator tests passed.
- All 24 downstream shapes passed accuracy and replay determinism checks against [CUDA FP8 FlashMLA](https://github.com/meituan-longcat/FlashMLA/tree/feature/ckv_fp8_per_token). Maximum output relative L2 error was 0.007768; maximum LSE absolute error was 2.862e-6.

Before/after suite comparisons retained the same pre-existing failures: 65/72 TLE tests and 10/11 selected NVIDIA lowering tests passed. Both cross-dtype MLIR files passed.

A separate non-TLE build and Compute Sanitizer were not run.

### DIFF
--- a/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/DotOpToLLVM/WGMMA.cpp
+++ b/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/DotOpToLLVM/WGMMA.cpp
@@ -216,7 +216,75 @@ LogicalResult convertDot(const LLVMTypeConverter *typeConverter,
   auto baseB = getOffsetedBase(loadedB, cast<MemDescType>(bTensorTy),
                                typeConverter, rewriter, loc);
   auto dShapePerCTA = getShapePerCTA(dTensorTy);
-  auto instrMNK = mmaEncoding.getInstrShape();
+  auto instructionMmaEncoding = mmaEncoding;
+#ifdef __TLE__
+  // An async accumulator chain may cross WGMMA operand types. Hopper's C
+  // register ownership depends on M/N and the warp/CTA distribution, not K.
+  // TLE keeps the preceding dot's encoding on the async C value, while a
+  // register-A operand carries the instruction shape required by the current
+  // dtype. Use that shape only when both encodings describe the same physical
+  // C register layout.
+  if (auto aDotEncoding =
+          dyn_cast<DotOperandEncodingAttr>(aTensorTy.getEncoding())) {
+    if (auto aMmaEncoding =
+            dyn_cast<NvidiaMmaEncodingAttr>(aDotEncoding.getParent())) {
+      if (aMmaEncoding != mmaEncoding) {
+        auto accInstrShape = mmaEncoding.getInstrShape();
+        auto currentInstrShape = aMmaEncoding.getInstrShape();
+        bool compatibleCLayout =
+            mmaEncoding.isHopper() && aMmaEncoding.isHopper() &&
+            accInstrShape[0] == currentInstrShape[0] &&
+            accInstrShape[1] == currentInstrShape[1] &&
+            mmaEncoding.getWarpsPerCTA() == aMmaEncoding.getWarpsPerCTA() &&
+            mmaEncoding.getCTALayout() == aMmaEncoding.getCTALayout();
+        if (compatibleCLayout) {
+          instructionMmaEncoding = aMmaEncoding;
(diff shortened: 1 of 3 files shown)


zeroherolin:
Thx to @July-h5kf3 

I have reviewed your code, and the fix looks correct: on H20 the 8 FP8/BF16 cross\-dtype chain variants that abort on base compile on the PR and match a torch fp32 reference, and lit, `test_core.py -k dot` and the 8 same\-dtype variants match base\. Two small items, neither blocks merging\.

## 1\. The register\-A check compares M/N, which do not affect the A layout

`WGMMA.cpp:279-280` rejects a register A whose parent `instrShape` M/N differ from the accumulator's, but the A layout only depends on kWidth, warpsPerCTA and CTALayout: `triton-tensor-layout` prints identical layouts for parent N=64 and N=128\. `marked_n`/`unmarked_n` therefore reject IR that base lowers correctly\. Keep the warpsPerCTA/CTALayout comparison, drop M/N and the two N tests\.

## 2\. The accumulator encoding no longer describes the instruction; say so in the code

After this PR a TLE chain carries `warp_group_dot` ops whose result `instrShape[2]` contradicts the operand dtype and whose register\-A parent differs from the result encoding\. Add a comment at `TleLowerWGMMA::materializeMMAAccumulator`: the accumulator keeps the first dot's encoding and K is re\-selected at lowering\. Cosmetic: the second split of `test_tle_cross_dtype_wgmma_accumulator.mlir` repeats the copyright header\.

July-h5kf3:
> Thx to @July-h5kf3
> 
> I have reviewed your code, and the fix looks correct: on H20 the 8 FP8/BF16 cross-dtype chain variants that abort on base compile on the PR and match a torch fp32 reference, and lit, `test_core.py -k dot` and the 8 same-dtype variants match base. Two small items, neither blocks merging.
> 
> ## 1. The register-A check compares M/N, which do not affect the A layout
> `WGMMA.cpp:279-280` rejects a register A whose parent `instrShape` M/N differ from the accumulator's, but the A layout only depends on kWidth, warpsPerCTA and CTALayout: `triton-tensor-layout` prints identical layouts for parent N=64 and N=128. `marked_n`/`unmarked_n` therefore reject IR that base lowers correctly. Keep the warpsPerCTA/CTALayout comparison, drop M/N and the two N tests.
> 
> ## 2. The accumulator encoding no longer describes the instruction; say so in the code
> After this PR a TLE chain carries `warp_group_dot` ops whose result `instrShape[2]` contradicts the operand dtype and whose register-A parent differs from the result encoding. Add a comment at `TleLowerWGMMA::materializeMMAAccumulator`: the accumulator keeps the first dot's encoding and K is re-selected at lowering. Cosmetic: the second split of `test_tle_cross_dtype_wgmma_accumulator.mlir` repeats the copyright header.

Done

July-h5kf3:
Fixed!

sunnycase (approved):
LGTM

PR #1018, opened 2026-08-19
[FIX][TLE]: fix dsa buffer turning into tensor when crossing for scopes

finneyyan:
Background: When the frontend enters a for/if region, it clones the scope and reconstructs variables using the IR handle via _unflatten_ir. The buffer_type of dsa does not implement this interface, resulting in buffers allocated outside the loop/branch being reconstructed as regular tl.tensor after cloning, leading to compilation errors.

Change: Complete the missing _unflatten_ir for buffer_type to ensure type stability after cloning; modify the test_bind_buffer.py test case to cover this scenario

### DIFF
--- a/python/test/tle/test_bind_buffer.py
+++ b/python/test/tle/test_bind_buffer.py
@@ -4,7 +4,7 @@ import triton.language as tl
 
 from triton.compiler.compiler import ASTSource
 from triton.compiler.code_generator import ast_to_ttir
-from triton._C.libtriton import ir, tle as tle_ir
+from triton._C.libtriton import ir, tle as tle_ir, buffer_ir
 from triton._C.libtriton.ascend import ir as ascend_ir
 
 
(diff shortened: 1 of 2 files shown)


python/test/tle/test_bind_buffer.py
```diff
@@ -23,19 +23,25 @@ def compile_kernel(kernel, signature, constants):
...
     ascend_ir.load_dialects(context)
     module = ast_to_ttir(kernel, src, context, Options(), {}, {})
     return str(module)
 
 
 @triton.jit
-def bind_buffer():
+def bind_buffer(XBLOCK: tl.constexpr, N: tl.constexpr):
+    # buffer 在 for 外 alloc、循环内 to_tensor: 进循环体触发 scope 克隆,
+    # 克隆后 buffer 必须仍是 buffer 而不是 tensor (buffer_type._unflatten_ir)
```
sunnycase: Please comment in english.

neoskypig:
@sunnycase help to review, thanks.

sunnycase (approved):
LGTM

PR #1020, opened 2026-08-20
[TLE] Fix default cluster_dims=(1,1,1) for backends that omit it

tengqm:
`CompiledKernel.__init__` builds the `KernelMetadata` namedtuple from the metadata dict emitted by the backend. Backends without cluster-launch support (hcu, mthreads) omit `cluster_dims`, so the field is missing from the namedtuple. `torch._inductor` reads `kernel.metadata.cluster_dims` unconditionally when a kernel carries metadata, raising `AttributeError` during KV-cache warmup — reproduced with vllm 0.24.0 on the Hygon DTK 26.04 runtime (flagtree 0.6.1+hcu3.6).

Default `cluster_dims` to `(1,1,1)` before building the namedtuple, the same value vendor triton and the iluvatar overlay already use. Backends that do emit the field are unaffected.

Fixing it at the source also lets the MUSA adapter in vllm-plugin-FL drop its `patch_inductor_triton_for_musa` monkey-patch (its comment marks it `TODO remove once flagtree adds cluster_dims`).

Applying the same one-line change in the container lets the vllm 0.24.0 F-path serve proceed past the AttributeError and reach application startup.


zhzhcookie (approved):
LGTM

PR #1021, opened 2026-08-20
[Feature] enflame: fix __nv_nearbyinf symbol do not support problem a…

chaaa-a:
fix __nv_nearbyinf symbol do not support problem in gcu400 and atomic operation bug


chaaa-a:
涉及到enflame版本重构，该pr放弃，不合入

PR #1023, opened 2026-08-21
fix(nvidia): require a real CUDA torch in is_active()

### LINKED ISSUE #1022: fix(nvidia): nvidia driver is_active() fires on Ascend NPU under torch_npu TORCH_TRANSFER_TO_NPU=1
### Bug description

`third_party/nvidia/backend/driver.py` `CUDABackend.is_active()` reports the nvidia backend active on machines that have **no NVIDIA GPU** when `torch_npu` is loaded with the transfer-to-NPU shim (`TORCH_TRANSFER_TO_NPU=1`):

```python
return torch.cuda.is_available() and (torch.version.hip is None)
```

On Ascend NPU hosts the shim makes `torch.cuda.is_available()` return `True` (it redirects CUDA API calls to the NPU), and the torch build is not a HIP build, so `torch.version.hip is None` also holds. The nvidia driver therefore reports active on a pure NPU host. The Ascend `NPUDriver.is_active()` (bisheng `npucompiler` probe) is *also* active on that host, so driver selection sees two active backends and `triton.runtime.driver._create_driver()` fails.

### Environment

- Ascend 910B, CANN 9.0.0, `torch_npu` with `TORCH_TRANSFER_TO_NPU=1`
- FlagTree runtime wheel (Triton-based), nvidia driver at `third_party/nvidia/backend/driver.py:876` (main @ 52678a8b7)

### Steps to reproduce

Run any Triton workload on an Ascend host with `TORCH_TRANSFER_TO_NPU=1` set (the standard torch_npu shim mode). Driver probing enumerates backends via `is_active()`; nvidia returns `True` (CUDA-available faked by the shim, not a HIP torch) while the ascend driver also returns `True`, so driver selection errors instead of resolving to `NPUDriver`.

### Expected behavior

`is_active()` should test for a *real* CUDA torch build, not just CUDA-API availability. A genuine CUDA build always carries a non-`None` `torch.version.cuda`; the shim does not fake it. On the Ascend host the check should return `False` (observed after the fix below: `amd=False, ascend=True, nvidia=False` → `NPUDriver` selected).

### Proposed fix

```python
return (
    torch.cuda.is_available()
    and torch.version.cuda is not None
    and (torch.version.hip is None)
)
```

This matches the guard upstream triton uses in its nvidia driver. No behavior change on real CUDA/HIP hosts (a CUDA torch always has `version.cuda` set).

tengqm:
## Target

`triton_v3.5.x` — the ascend backend lives on the triton 3.5 line, not `main`; the two-active-driver conflict this fixes only materializes there.

## What does this PR do?

Fixes `CUDABackend.is_active()` in `third_party/nvidia/backend/driver.py` firing on Ascend NPU hosts under torch_npu's transfer-to-NPU shim.

The current check — `torch.cuda.is_available() and (torch.version.hip is None)` — treats CUDA-API *availability* as evidence of a real CUDA torch. On Ascend hosts the shim (`TORCH_TRANSFER_TO_NPU=1`) fakes `torch.cuda.is_available()`, and the torch build is not a HIP build, so the nvidia backend reports active on a machine with no NVIDIA GPU. The Ascend `NPUDriver` is also active, so `triton.runtime.driver` sees two active backends and selection fails.

The fix requires `torch.version.cuda is not None` — a genuine CUDA build always carries the CUDA version string, and the shim does not fake it (this is the same guard upstream triton uses).

## Issue

Fixes #1022

## Verification

- Ascend 910B (CANN 9.0.0, torch_npu + `TORCH_TRANSFER_TO_NPU=1`): after the change driver probing yields `amd=False, ascend=True, nvidia=False`, and driver resolution lands on `NPUDriver`; the Triton workload runs end-to-end
- Real CUDA / HIP hosts: a CUDA torch always has `version.cuda` set and `version.hip` None — no behavior change
- `python -c "import ast; ast.parse(...)"` clean; change is a single boolean expression

### DIFF
--- a/third_party/nvidia/backend/driver.py
+++ b/third_party/nvidia/backend/driver.py
@@ -762,7 +762,10 @@ class CudaDriver(GPUDriver):
     def is_active():
         try:
             import torch
-            return torch.cuda.is_available() and (torch.version.hip is None)
+            # `torch.cuda.is_available()` can be faked by accelerator shims
+            # (torch_npu with TORCH_TRANSFER_TO_NPU=1 reports the NPU as CUDA);
+            # a real CUDA build always carries a non-None `torch.version.cuda`.
+            return (torch.cuda.is_available() and torch.version.cuda is not None and (torch.version.hip is None))
         except ImportError:
             return False
 


zhzhcookie:
`ascend` backend is not on main branch.

tengqm:
Good catch — the ascend backend lives on the triton 3.5 line, not main. I have rebased this PR onto `triton_v3.5.x` (squashed to a single commit, diff unchanged: one hunk in the nvidia `is_active()`) and retargeted it there, where the two-active-driver conflict is actually reproducible. Thanks!

PR #1025, opened 2026-08-21
[Ascend] make testing.py torch imports lazy

### LINKED ISSUE #1024: ascend: importing triton before torch crashes with Failed to load the backend extension: torch_npu
## Problem

On Ascend hosts, `import triton` executed before (or independently of) `import torch` crashes with:

```
RuntimeError: Failed to load the backend extension: torch_npu. You can disable extension auto-loading with TORCH_DEVICE_BACKEND_AUTOLOAD=0.
```

This hits every triton-first entry point: the `compiler triton` switch, `do_bench` invocations, and any script that imports triton directly without importing torch first.

## Root cause

`import triton` runs module-level `_discover_backends()` in `python/triton/backends/__init__.py`, which enumerates entry points and loads `triton.backends.ascend`. The ascend backend package's `__init__.py` imports `do_bench_npu` from `testing.py`, and `testing.py` has **top-level** `import torch; import torch_npu` (lines 27-28).

On the first-ever torch import, `torch/__init__.py` `_import_device_backends()` attempts to autoload the `torch_npu` extension and can fail (e.g. when the extension is not yet on the path or the environment is not fully initialized), raising the error above. Because the failure happens during triton's backend discovery, the entire `import triton` crashes.

Importing torch first masks the issue: torch_npu is already registered in `sys.modules` by then, so `testing.py`'s import hits the cache and never triggers the autoload path.

## Fix

Move `import torch` / `import torch_npu` from the module top level into the two functions that actually use them (`do_bench_npu_profiler`, `do_bench_npu_mspti`). `testing.py` has no module-level torch/torch_npu usage — every use is inside a function. This matches the method-level lazy-import style already used in the rest of this backend (`driver.py`, `utils.py`, `backend_register.py`).

## Impact

Non-blocking robustness fix: megatron E2E runs are unaffected (they always import torch first), but standalone triton users on Ascend are broken today.

tengqm:
## Summary

Make `third_party/ascend/backend/testing.py`'s torch imports lazy. Currently the module has top-level `import torch; import torch_npu`, so `import triton` (which discovers backends and loads `triton.backends.ascend` -> `testing.py`) can crash on the first torch import when torch's device-backend autoload of `torch_npu` fails:

```
RuntimeError: Failed to load the backend extension: torch_npu. You can disable extension auto-loading with TORCH_DEVICE_BACKEND_AUTOLOAD=0.
```

Every triton-first entry point (the `compiler triton` switch, `do_bench`, standalone scripts) is affected. Importing torch first masks the bug by populating `sys.modules` before the autoload path can run.

## Change

- Remove the top-level `import torch` / `import torch_npu` from `testing.py`
- Add both imports inside `do_bench_npu_profiler` and `do_bench_npu_mspti` — the only two functions that use torch/torch_npu

All torch/torch_npu usage in this file is already inside function bodies; there is no module-level use. This matches the method-level lazy-import style already used elsewhere in this backend (`driver.py`, `utils.py`, `backend_register.py`).

Verified: `import triton` no longer imports torch (a module-level-import blocker confirms `testing.py` loads clean without torch); `py_compile` passes.

Fixes #1024


zhzhcookie (approved):
LGTM

zhzhcookie:
Closes https://github.com/flagos-ai/FlagTree/issues/1024

PR #976, opened 2026-08-13
[TLE][PPU] Add INT8 AIU support to PPU0010 V1 lowering

July-h5kf3:
## Background

TLE promotes `tle.load(block_ptr, is_async=True)` to PPU AIU asynchronous copies. The FP16/BF16 path already lowers through PPU0010 lowering V1, but the INT8 path did not have a complete B8 AIU data path.

This is not a PPU0010 hardware limitation. The public ACTLIZE implementation provides direct evidence:

- [ACTLIZE 1.0.0 README](https://github.com/t-head/actlize/blob/6a0c8748406dfab412220134baee0575ba980818/README.md#L3-L68) lists INT8 GEMM, AIU load, and TSM swizzle load support for PPU 1.0, and includes ZW810E in the supported hardware list.
- [copy_ppu0010_aiu.hpp](https://github.com/t-head/actlize/blob/6a0c8748406dfab412220134baee0575ba980818/include/cute/arch/copy_ppu0010_aiu.hpp#L191-L255) implements both non-transposed and transposed 8-bit linear/swizzled AIU loads for PPU0010 and emits `ppu.cp.async.aiu.bulk.tensor.shared.global.padz.swzl.2d.b8`.
- [default_gemm_configuration.hpp](https://github.com/t-head/actlize/blob/6a0c8748406dfab412220134baee0575ba980818/examples/15_ppu_cute_gemm/default_gemm_configuration.hpp#L253-L294) configures PPU0010 INT8 GEMM with `MainloopPPUAiu<3>`, `DefaultGemm_AIU_Operand<int8_t, ...>`, and S8 x S8 -> S32 tensor-core MMA.

Moving PPU0010 to lowering V2 is not a valid workaround because V2 emits the unsupported `2d.tile` instruction form. This PR keeps the existing V1 dispatch and completes B8 support in V1.

## Root cause

The PPU0010 V1 lowering contained several related FP16 assumptions:

1. V1 async-copy lowering hard-coded the `.b16` suffix instead of selecting the instruction from the source element width.
2. `getPPUAIUV1SwizzledSharedPtrs` hard-coded FP16 geometry for the 32-byte slice, 16-byte vector, and 128-byte swizzle period.
3. Shared-to-dot lowering only used the B16 `ldmatrix` form and did not emit the native B8 TSM swizzle load.
4. For K >= 64, the shared pointer for the second K32 fragment omitted `slice_id * cubeW * 32` INT8 elements.
[...]

PR #1026, opened 2026-08-21
[TLE][PPU] Add INT8 AIU support to PPU

tianyi-dialect:
Instead of https://github.com/flagos-ai/FlagTree/pull/976


zhzhcookie (approved):
LGTM

PR #1033, opened 2026-08-21
[TLE][AMD] Rematerialize TLE matmul loads after WMMA selection

zihaomu:
## What

Re-run the TLE local-pointer load optimization once more after the AMD WMMA /
matmul layout selection pass (`add_accelerate_matmul`) in `make_ttgir`.

## Why

For tiled TLE matmul, operands are staged through shared memory via
`tle.gpu.local_ptr`. Before layout selection, the local-pointer load
optimization runs, but WMMA layout selection then re-introduces a second
staging of the same operands (an extra `local_load` reload). Re-running
`add_optimize_local_pointer_loads` after `add_accelerate_matmul` rematerializes
those loads so operands stay in registers instead of being reloaded from LDS.

The change is gated on `tle is not None`, so non-TLE compilation and other
backends are unaffected.

## Testing (gfx1201 / RDNA4)

- New unit test `test_local_pointer_tiled_matmul_amd_avoids_second_staging`
  asserts the post-optimization TTGIR keeps two `ttg.local_alloc` / two
  `ttg.local_store` for the two K tiles, has no `ttg.local_load` reload,
  uses 4096B shared, and emits `v_wmma`.
- The test targets native AMD only, guarded by
  `is_hip() and not FLAGTREE_BACKEND` (excludes HCU), matching the merged
  identity convention.
- `python/test/tle/unit/test_tle_gpu_local_ptr.py`: 13 passed.
- `test_tle_cumsum.py + test_tle_gpu_local_ptr.py`: 23 passed, 2 expected skips.
- Ruff 0.9.1 and YAPF 0.43.0: clean.
- Rebased onto current upstream `main`: conflict-free, 0 commits behind.

`compiler.py` is a pure Python pass-ordering change; no native rebuild required.

### DIFF
--- a/python/test/tle/unit/test_tle_gpu_local_ptr.py
+++ b/python/test/tle/unit/test_tle_gpu_local_ptr.py
@@ -11,6 +11,8 @@ import torch
 import triton
 import triton.language as tl
 import triton.experimental.tle.language as tle
+from triton._flagtree_backend import FLAGTREE_BACKEND
+from triton._internal_testing import is_hip
 
 BLOCK_SIZE = 64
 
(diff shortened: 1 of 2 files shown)


sunnycase (approved):
LGTM

sunnycase:
@zihaomu Please fix the conflicts.

zihaomu:
@sunnycase Fixed

PR #1037, opened 2026-08-23
[TLE][MTHREADS] Support TLE set_layout interface on musa backend

Kylin1207:
### tle.gpu.set_layout(x, layout)

  Applies a compile-time distributed layout constraint to a tensor while preserving its shape, element type, and numerical value.

  Supported layouts include:

  - Blocked and Slice encodings.
  - Backend-local MusaWmmaEncoding.
  - Backend-local MusaSqmmaEncoding.
  - Backend-local MusaDotOperandEncoding.

  The target layout must be a distributed encoding and its rank must match the source tensor rank. Shared-memory layouts and incompatible encodings are rejected with an explicit diagnostic.

  ### Layout propagation

  The requested layout is treated as a hard constraint and propagated through:

  - Elementwise, broadcast, reshape, expand-dims, reduction, and scan operations.
  - Structured and unstructured control flow, including block arguments and yielded values.
  - Pointer-producing expressions and memory operations.
  - Tensor descriptor loads across Coalesce and explicit-layout finalization.
  - Private/noinline function arguments, return values, call operands, and call results.
  - Multi-level call graphs using a module-level fixed point independent of caller/callee declaration order.

  Shared layout-polymorphic producers are isolated when different users require incompatible Slice layouts. Safe register producers may be cloned by layout domain, while memory roots such as loads are not duplicated.

  Conflicting hard layouts are rejected instead of being silently overwritten.

  ### Lowering path

  - The frontend dispatches to the mthreads specialization under triton.experimental.tle.language.gpu.mthreads.
  - The operation is represented by the backend-local musa_tle.set_layout op; the shared TLE dialect is not linked into the mthreads build.
  - The MUSATLE verifier validates the distributed encoding and tensor rank before layout propagation.
  - During TTIR → TTGIR conversion, the requested encoding is materialized directly in tensor types.
  - musa_tle.set_layout is removed after propagation and does not remain as a runtime operation.
  - Backend-local explicit-result metadata preserves hard descriptor layouts through Coalesce and is validated and removed by the finalizer.
  - Existing Triton function, call, and return conversion patterns consume the synchronized tensor types without changing the non-TLE path.

  ### Limitations

  - Source and result must be ranked tensors with the same shape and element type.
  - The target must be a supported distributed encoding with matching rank.
  - Shared-memory encodings cannot be used as the target layout.
  - Two incompatible hard layout requirements in the same propagation domain are rejected.
  - Explicit tensor layout ABIs on external tt.func declarations are not supported; private/noinline helpers are supported.

### DIFF
--- a/python/triton/experimental/tle/language/gpu/mthreads/__init__.py
+++ b/python/triton/experimental/tle/language/gpu/mthreads/__init__.py
@@ -19,5 +19,16 @@
 # SOFTWARE.
 
 from . import buffer, common, copy, pipe, warp_specialize, wgmma
+from .types import MusaDotOperandEncoding, MusaSqmmaEncoding, MusaWmmaEncoding
 
-__all__ = ["buffer", "common", "copy", "pipe", "warp_specialize", "wgmma"]
+__all__ = [
+    "buffer",
+    "common",
+    "copy",
+    "pipe",
+    "warp_specialize",
+    "wgmma",
+    "MusaDotOperandEncoding",
+    "MusaSqmmaEncoding",
+    "MusaWmmaEncoding",
+]
--- a/python/triton/experimental/tle/language/gpu/mthreads/layout.py
+++ b/python/triton/experimental/tle/language/gpu/mthreads/layout.py
@@ -1,0 +1,204 @@
+# Copyright 2025-     FlagOS Contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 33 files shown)


sunnycase:
Thanks for this work! Could you please follow the pattern used in `https://github.com/flagos-ai/FlagTree/blob/main/python/triton/experimental/tle/language/gpu/__init__.py` and move the MUSA encoding classes into `mthreads/types.py`, then re-export them from `mthreads/__init__.py`? This would let users access them directly as `tle.gpu.mthreads.MusaDotOperandEncoding`. Thank you!

Kylin1207:
Thanks for the suggestion. I’ve moved the MUSA encoding classes to mthreads/types.py and re-exported them from
  mthreads/__init__.py. They are now accessible as tle.gpu.mthreads.MusaDotOperandEncoding and the corresponding
  WMMA/SQMMA encodings.

sunnycase (approved):
LGTM

PR #1039, opened 2026-08-24
[TLE][AMD] gfx1201 TopK provider selection + tuning

zihaomu:
## Summary

Adds gfx1201 (RDNA4) provider selection **and** launch-config tuning for the TLE TopK tutorial, plus a self-contained benchmark harness that produces the evidence for both. All behavior is gated to `hip`/`gfx1201`; other backends and archs are unchanged.

### Why one PR (not two)

The selector threshold and the kernel tuning are coupled: tuning the radix path shifts the radix-vs-streaming crossover, so the selector threshold has to be calibrated against the *tuned* kernels. Splitting would either (a) land a threshold that the tuning PR immediately rewrites, or (b) land a threshold that is miscalibrated until the tuning lands. Keeping them together puts the dispatch decision next to the tuning that justifies it and the benchmark that proves it. The three concerns are separated into three reviewable commits.

## Commits

1. **Add gfx1201 TopK provider selector** — `_topk_provider(n_cols, k)` + `topk(...)` dispatch wrapper; `run_correctness` validates the selected path. Non-gfx1201 targets always use radix (unchanged behavior). Adds a fast, GPU-free boundary unit test.
2. **Tune gfx1201 TopK launch configs and recalibrate selector** — `_radix_launch_config` (RADIX_BITS=8, wider BLOCK_N / more warps for large N) and `_triton_launch_config` (more warps for tall-narrow, large-k); selector threshold recalibrated to the tuned crossover.
3. **Add self-contained gfx1201 TopK benchmark harness** — reproducible radix/triton/torch A/B benchmark emitting selected provider, per-provider launch configs, p50 quantiles with a bootstrap speedup CI, and the run environment as JSON. Timing/stats/environment helpers are inlined so it does not pull in the general AMD microbenchmark module.

## Selector

Only on `gfx1201`:

```python
if (n_cols <= 8192 and k <= 8) or (n_cols <= 2048 and k <= 32):
    return "triton"
return "radix"
```

Boundary table (from the unit test): `(128,8)`, `(8192,8)`, `(2048,32)` → triton; `(8193,8)`, `(8192,32)`, `(2049,32)`, `(512,64)`, `(65536,2)` → radix. `(hip,gfx942)` and `(cuda,sm90)` → radix (unchanged).

## Tuning evidence (do_bench, M=8, fp16, tuned vs untuned radix)

| shape (n_cols, k) | speedup |
|---|---|
| (8192, 32)   | +27% |
| (16384, 32)  | +31% |
| (32768, 64)  | +33% |
| (65535, 32)  | +18% |
| (65536, 2)   | +37% |
| (131072, 8)  | +38% |

Reproduce:

```
python python/benchmarks/tle/benchmark_topk_amd.py --shape all --dtype float16 --output topk_gfx1201.json
```

## Testing

- `python/test/tle/unit/test_tle_topk_selector.py` — 2 passed (boundary + other-target preservation)
- `python/tutorials/tle/03-topk.py --only_unit_test` — correctness passed (radix + triton + selected path)
- benchmark harness — end-to-end run produces valid JSON (`arch: gfx1201`, `selected_provider`, launch configs, `radix_bits: 8`)
- ruff 0.9.1 + yapf 0.43.0 clean

Rebased onto latest `upstream/main` (`ddbecaa3b`).


sunnycase (approved):
LGTM

PR #1043, opened 2026-08-25
[Nvidia][TritonGPU] Support interleaved concat, heterogeneous segments, and segmented merging

zeroherolin:
Completes the Concat Dot Operand model started in #892. That PR handled one shape: a user-written join tree, laid end to end, folded into a dot operand. This PR adds the shapes that arrive **already segmented**, the fragments that arrive **interleaved**, and the chains whose segments are **unequal**.

Two internal ops carry the model. `ttg.concat_dot_operand` states that N fragments form a wider operand; `ttg.extract_dot_operand` states that a value is slice *i* of one. Every entry form is normalised into them, then proven: the K segments cover the wide operand seamlessly, and the wide layout differs from each fragment’s in register bases alone. Merging a dot chain additionally requires the segments to share one numerical contract with no ordering point between them. The proofs read types rather than IR, so a rejected candidate leaves no IR to clean up.

## Scenarios

**1. Join tree** (#892) — matched before layout assignment, so the encoding is chosen for one wide operand rather than for a rank-5 join chain.

```python
a = tl.join(tl.join(a0, a1), tl.join(a2, a3)).permute(0, 3, 2, 1).reshape((M, K))
c = tl.dot(a, b)
# -> one ttg.concat_dot_operand over 4 fragments, one dot
```

A tree built in stages (stack pairs, then stack the results) is flattened into the same single op.

**2. Split tree** — the dual, and the producer the merge path needs.

```python
a4 = a.reshape(M, 2, 2, FRAG).permute(0, 3, 2, 1)
aL, aR = tl.split(a4); a0, a1 = tl.split(aL); a2, a3 = tl.split(aR)
# -> ordered ttg.extract_dot_operand off the wide operand
```

Which slice a leaf holds comes from axis significance, not split order: `reshape` factors K most-significant-first, but splits peel the trailing axis first. `permute(0,3,2,1)` gives leaf order `0,1,2,3`; `permute(0,3,1,2)` gives `0,2,1,3`.

**3. MMAv2 register/register** — both operands are register extracts, so both need a complete cover. Axes differ: A slices `dim=1`, B slices `dim=0`.

```mlir
%a0, %a1 = ttg.extract_dot_operand %aw {dim = 1, index = 0/1}
%b0, %b1 = ttg.extract_dot_operand %bw {dim = 0, index = 0/1}
%d0 = tt.dot %a0, %b0, %acc
%d1 = tt.dot %a1, %b1, %d0
// -> %d = tt.dot %aw, %bw, %acc
```

**4. Segment count and size** — neither the matcher nor the K-cover proof requires a power of two or equal segments. The verifier sums per-fragment extents instead of assuming one common extent; the register-map proof constrains extents, not their count.

Two unequal powers of two never sum to a power of two, so the smallest heterogeneous case the layout verifier admits needs three fragments, e.g. `32 + 32 + 64`.

Unequal extents make the cover check stricter: `concat` of ordered extracts folds back to its source only when each extract starts exactly where the previous ended, not merely when the indices ascend. `concat(extract(w,0,32), extract(w,1,16), extract(w,2,16))` reads `w[0:32] w[16:32] w[32:48]` — it still matches `w`'s type, but folding it would duplicate one slice and drop another.

**5. Interleaved fragments** — the same join tree with the permute moved *behind* the in-fragment K coordinate. Sub-byte packing produces this directly: one byte holds elements `2j` and `2j+1`.

```python
lo = ((packed       & 0xF) * scale_lo).to(tl.float16)
hi = (((packed >> 4) & 0xF) * scale_hi).to(tl.float16)
w  = tl.reshape(tl.permute(tl.join(lo, hi), (0, 2, 1)), (BLOCK_K, BLOCK_N))
```

`permute(0,2,1)` states `k = kIn * numFrags + i` instead of `k = i * extent + kIn`; the `interleaved` unit attribute selects that map. On a trailing `dim` it is the identity permutation, so a join tree with no transpose at all is also matched.

Interleaved operands are always expanded back to a join tree, never relabeled: a `dot_op` thread owns its K elements in contiguous runs, so alternating fragments every element puts the data a thread needs on other threads. The op is still worth forming, because for as long as it exists layout propagation sees one wide operand and picks the encoding for it.

Nested interleaving is left alone — stacking pairs then stacking the results gives `a0 c0 b0 d0`, not the `a0 b0 c0 d0` of a four-way interleave.

## Not merged

| Rejection | Why |
|---|---|
| Multiply-used or late-defined fragment | a fragment feeding anything but its own dot stays live, and one defined after the chain head would pull the merged dot past its producer |
| Incomplete cover | the slices leave a gap, an overlap, or a padded tail |
| Unrepresentable merged shape | `DotOperand` is power-of-two LinearLayout only, so a merged K of 80 stays segmented |
| Memory effect between two dots | read timing becomes semantically significant |
| Async ordering point between two dots | a WGMMA wait/commit, a token-bearing `local_load`, or a TLE shared-operand fence is an ordering boundary that no memory effect models |
| Discardable scheduling attributes | e.g. pipeline stage; merging would drop the contract |
| Chain disagreeing on `inputPrecision` or `maxNumImpreciseAcc` | the segments do not share one numerical contract |

`tt.dot_scaled` and MMAv5/TMEM are not rejected by a check — the matcher keys on `tt.dot`, so they never enter it. Extending to either needs a cover proof for the scale tensor and an operand model for TMEM respectively, not a relaxed predicate.

A merged K is not re-checked against `maxNumImpreciseAcc`: the attribute only drives chunked accumulation in the WGMMA lowering, and the matcher keys on MMAv2, so widening K introduces no new accumulation boundary there.

Operands staged through shared memory do reach the concat; the expand pass rebuilds the join tree for them once the layouts are final. Merging segments that are already `memdesc_subslice` views of one buffer is a separate shape that no source-level kernel produces, and `optimize-dot-operands` already covers it.

## Pipeline placement

| Pass | Runs | Because |
|---|---|---|
| `tritongpu-concat-dot-operand` | before layout assignment | otherwise encodings are picked for the join chain, and the wide operand is only reachable through a reshape between two unrelated layouts |
| `tritongpu-merge-segmented-dot` | after `accelerate-matmul`, and **again** after `optimize-dot-operands` | the match needs the mma layout; the second run catches chains that only appear once register layouts settle |
| `tritongpu-expand-concat-dot-operand` | last | whether the relabel is possible follows from the final encoding |

The expand pass closes with a verification walk: an op that can neither be relabeled nor expanded is reported there, naming both why the relabel failed and why the expansion could not serve it, rather than failing later in `ViewOpToLLVM`.

## Validation

- `concat-dot-operand.mlir` — 25 cases over four RUN lines (recognition, recognition + expand round trip, merging, split/merge interaction). Negative cases: out-of-order permute, `allowReorder` reshape, producer escape, accumulator-only reach, nested interleave, overlapping extracts.
- `invalid.mlir` — 7 new verifier cases, bringing the two ops to 11 in total: fragment type and non-contraction dim mismatch, result extent against the sum of the segments, interleave fragment equality, extract index range, minimum slice count, and slice divisibility.
- Every pattern was driven from Triton kernels reduced from the FlagGems operators that spell these shapes (`fused_marlin_moe`, `flash_mla`, `_weight_int4pack_mm_with_scales_and_zeros`), checked for numerics and against the TTGIR they compile to, on H20/sm90.

Everything stays behind `__FLAGTREE_CONCAT_DOT_OPERAND__`, defined only when `FLAGTREE_BACKEND` is unset; the pass source is conditionally compiled, so non-NVIDIA builds do not see it.


sunnycase (approved):
LGTM

PR #1052, opened 2026-08-26
[VERSION] Export triton_key from triton.compiler.compiler to align with the Triton 3.4 interface

tengqm:
torch 2.8 inductor (`torch/_inductor/codecache.py`, `CacheBase.get_system`) imports `triton_key` from `triton.compiler.compiler` to include the triton revision in its compile-cache key. FlagTree defines `triton_key` in `triton.runtime.cache` but never re-exports it on the compiler module, so `import torch._inductor.codecache` — and any `torch.compile` path — raises `ImportError` and aborts.

This re-exports the existing symbol on the compiler module. No behavior change.

Repro before this change:

    python3 -c "from triton.compiler.compiler import triton_key"
    # ImportError: cannot import name 'triton_key' from 'triton.compiler.compiler'

### DIFF
--- a/python/triton/compiler/compiler.py
+++ b/python/triton/compiler/compiler.py
@@ -29,7 +29,7 @@ from ..backends.compiler import Language
 from ..backends.compiler import BaseBackend, GPUTarget
 from .. import __version__, knobs
 from ..runtime.autotuner import OutOfResources
-from ..runtime.cache import get_cache_manager, get_dump_manager, get_override_manager, get_cache_key, triton_key
+from ..runtime.cache import get_cache_manager, get_dump_manager, get_override_manager, get_cache_key, triton_key as triton_key
 from ..runtime.driver import driver
 from ..tools.disasm import get_sass
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tengqm:
The pre-commit failure is a ruff F401 (unused import) auto-fix reverting the PR's only change — not a functional problem.

**Root cause:** ruff v0.9.1 runs with `--fix --exit-non-zero-on-fix`. The new `triton_key` is imported into `compiler.py` solely to be re-exported for external consumers (`from triton.compiler.compiler import triton_key`) — it is never used inside `compiler.py` itself, so ruff flags it as unused (F401) and its auto-fix deletes it, which is exactly the line this PR adds.

**Fix (one line):** make the re-export explicit so ruff recognizes it as intentional:

```python
from ..runtime.cache import get_cache_manager, get_dump_manager, get_override_manager, get_cache_key, triton_key as triton_key
```

`from x import y as y` is ruff's documented explicit re-export form and does not trigger F401. Alternatively `# noqa: F401` on the line works, but the `as` form is cleaner (no lint suppression).

The auto-commit push-back failure (`fatal: invalid reference: fix/metax-triton-key-reexport`) is a secondary symptom: once ruff has nothing to fix, no auto-commit is triggered, so that error disappears on its own.

All unit tests (including metax36x-unit-test) are green; this is purely the format gate.

zhzhcookie (approved):
LGTM

PR #1053, opened 2026-08-26
[XPU] Stop trusting Discrete offsets whose chain can wrap

Ason93:
OffsetAnalysis mocks program ids over a small range, so a `remsi` by a large constant in the offset chain never wraps during the analysis. For a gather offset such as `(offs % SRC_ROW) // INNER` the mocked samples look like a well-behaved Discrete access, whose invariant is that every lane of a core lies within [base, base + numElems) of that core's first lane.

UnrollControl::findDiscretePtrChain relies on that invariant and rewrites the gather into a contiguous gm2lm plus `lmPtr[offset - offset0]`. At the real wrap the offset jumps backwards by the modulus, the LM index goes far out of range and the kernel traps with a hardware exception.

Fall back to Unknown when the offset chain of a Discrete gm2lm contains a modulo by a constant larger than numElems, so the safe per-element gather path is used. Add a regression test that reproduces the index_copy-style scatter.

### DIFF
--- a/third_party/xpu/lib/Dialect/TritonXPU/Transforms/OffsetAnalysis.cpp
+++ b/third_party/xpu/lib/Dialect/TritonXPU/Transforms/OffsetAnalysis.cpp
@@ -1128,6 +1128,52 @@ public:
           memoryStateTransfer(memoryState, allOffsetStateResult[token]);
     }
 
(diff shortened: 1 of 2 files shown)


zhzhcookie (approved):
LGTM

PR #1063, opened 2026-08-28
[BUILD] Publish release packages to FlagOS Nexus via the shared build-infra workflow

shiptux:
## Why

`build-deb.yml` / `build-rpm.yml` (#607 / #794 / #796) produce the FlagTree deb/rpm, but nothing publishes them; FlagOS Nexus holds no FlagTree package.

## What

Adds `.github/workflows/upload-nexus.yml`, a thin caller of the shared [`flagos-ai/build-infra/.github/workflows/upload-nexus.yml`](https://github.com/flagos-ai/build-infra/blob/main/.github/workflows/upload-nexus.yml) — identical to the one libtriton_jit uses. Tag patterns match the build workflows (`[0-9]*`, `v[0-9]*`). The shared workflow:

- skips pre-release tags (`rc`/`a`/`b`/`dev`) on `push`, always publishes on `workflow_dispatch`
- waits for `build-deb.yml` (required) and `build-rpm.yml` (optional) of the same commit, downloads their artifacts
- uploads deb → `flagos-apt-hosted`, rpm → `flagos-yum-hosted` using the org-level `NEXUS_TOKEN`
- fails the job if packages were found but none uploaded

Note: because release tags are pushed with `GITHUB_TOKEN` from the delivery workflows, neither the build workflows nor this one fire on tags today (#1062). Until that is fixed, publishing is: dispatch `build-deb`/`build-rpm` on the tag, then dispatch this workflow with the build `run_id`.

Part of FEP-0019 Wave 1 (flagos-ai/community#59).

PR #1064, opened 2026-08-28
[CD] Dispatch build-deb/build-rpm after pushing a release tag

### LINKED ISSUE #1062: [BUILD] Release tags never trigger build-deb/build-rpm: tags are pushed with GITHUB_TOKEN by the delivery workflows
### Describe the bug

`build-deb.yml` / `build-rpm.yml` are meant to run on every release tag (`[0-9]*` and `v[0-9]*`, with `+backend` tags skipped — #796). In practice **no tag has ever triggered them**: `0.6.1` (2026-08-13), `0.6.1+*`, `0.6.2a1` (2026-08-25) all show zero packaging runs, and the only tag-event runs on record are `skipped` ones for `+backend` tags.

### Cause

All release tags are created by `flagtree-bot` inside the `*-delivery.yml` workflows, which push them with the workflow's own `GITHUB_TOKEN`:

```yaml
env:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
...
git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${{ github.repository }}.git
git push --force origin "${TAG}"
```

GitHub deliberately does not fire `push` (or any other) events for refs created with `GITHUB_TOKEN`, to prevent recursive workflow runs ([docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)). So the trigger pattern in the packaging workflows is correct but can never match; the glob fix in #796 could not have helped.

Verified with `gh api repos/flagos-ai/FlagTree/actions/runs?branch=0.6.1` → empty, and the tagger of `0.6.1` / `0.6.2a1` is `flagtree-bot` (`git for-each-ref --format='%(taggername)'`).

### Options

1. **Dispatch from the delivery workflow** — after the tag is pushed, call `gh workflow run build-deb.yml --ref "$TAG"` (and `build-rpm.yml`). Smallest change; keeps `GITHUB_TOKEN`; the packaging workflows already accept `workflow_dispatch`.
2. **Push the tag with a PAT or GitHub App token** instead of `GITHUB_TOKEN`, so the normal `push: tags:` trigger fires. Also makes any other tag-triggered workflow work, but needs a new secret and broader permissions.
3. Make the packaging jobs part of the delivery workflow itself (`workflow_call`). Cleanest long-term, largest diff.

I'd suggest (1) now and (3) when the delivery workflows are next refactored. Happy to send the PR for (1).

This also matters for FEP-0019 (flagos-ai/community#59): the FlagTree deb/rpm on FlagOS Nexus can only be produced by manual `workflow_dispatch` until this is fixed.

shiptux:
## Why

Release tags are created inside the `*-delivery.yml` workflows and pushed with the job's `GITHUB_TOKEN`. GitHub deliberately fires no `push` event for refs created with that token (to prevent recursive runs), so `build-deb.yml` / `build-rpm.yml` — which trigger on `[0-9]*` / `v[0-9]*` tags — have never run for a release tag: `0.6.1`, `0.6.2a1` and every other tag show zero packaging runs. Details and verification in #1062. The same applies to the Nexus upload caller from #1063: it triggers on tag push, so it would never fire for delivery-created tags either.

## What

`workflow_dispatch` is the documented exception to that rule, so each delivery workflow now dispatches the packaging chain on the tag it just pushed:

- one new step right after **FlagTree push tag**, identical in all 25 `*-delivery.yml` files (same anchor block in every file)
- dispatches `build-deb.yml`, `build-rpm.yml` and `upload-nexus.yml`; the upload is skipped (not fatal) while #1063 hasn't landed on the tagged commit
- plain `curl` against `POST /repos/{repo}/actions/workflows/{wf}/dispatches` — no dependency on `gh` being installed on the self-hosted runners
- **only stable release tags are dispatched**: because dispatch bypasses the upload workflow's own tag guard, the step applies the same stable-tag allowlist regex as that guard (`^v?[0-9]+(\.[0-9]+){0,3}(\.post[0-9]+)?$`). This skips backend-variant tags (`<version>+<backend>`) and also plain pre-release tags (`0.6.0rc1`), which a bare `+`-check would let through
- `permissions: actions: write` added next to the existing `contents: write`, which the dispatch call requires

The packaging workflows themselves are unchanged; on a dispatched run they build the tag's commit exactly as a tag push would have. On the build-infra side, flagos-ai/build-infra#912 makes the dispatched upload wait for this exact commit's builds instead of falling back to the latest successful run.

## Alternatives considered

Pushing the tag with a PAT / GitHub App token would make the normal `push` trigger work but needs a new secret with broader permissions; folding packaging into the delivery workflows via `workflow_call` is cleaner long-term but a much larger change. This is the minimal fix; happy to follow up with either if preferred.

Fixes #1062. Related: #1063 (publishing the built packages to FlagOS Nexus), flagos-ai/build-infra#912 (build-wait on dispatched uploads).

### DIFF
--- a/.github/workflows/aipu3.3-delivery.yml
+++ b/.github/workflows/aipu3.3-delivery.yml
@@ -22,6 +22,7 @@ name: CD-AIPU3.3
 
 permissions:
   contents: write
+  actions: write
 
 on:
   workflow_dispatch:
@@ -168,6 +169,27 @@ jobs:
             echo "git push failed after $MAX_RETRY attempts"
             exit 1
           fi
+
+      - name: Trigger packaging workflows for the release tag
+        # Tags pushed with GITHUB_TOKEN never fire `push` events (GitHub blocks
+        # recursive workflow runs), so build-deb/build-rpm would otherwise stay
+        # idle for every release tag (#1062). workflow_dispatch is the documented
+        # exception, so dispatch them explicitly. Backend-variant tags
+        # (<version>+<backend>) are not packaged and are skipped here as well.
+        if: ${{ !contains(env.FLAGTREE_WHEEL_VERSION, '+') }}
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG="${FLAGTREE_WHEEL_VERSION}"
+          for wf in build-deb.yml build-rpm.yml; do
+            echo "Dispatching ${wf} on ${TAG}"
+            curl -fsS -X POST \
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+              "https://api.github.com/repos/${{ github.repository }}/actions/workflows/${wf}/dispatches" \
+              -d "{\"ref\":\"${TAG}\"}"
+          done
 
       - name: FlagTree upload whl
         shell: bash
--- a/.github/workflows/amd3.6-delivery.yml
+++ b/.github/workflows/amd3.6-delivery.yml
@@ -22,6 +22,7 @@ name: CD-Amd3.6
 
 permissions:
   contents: write
+  actions: write
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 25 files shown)


shiptux:
Pushed a second commit: the dispatch step now applies the same stable-tag allowlist as the shared upload workflow's guard (plain pre-release tags like `0.6.0rc1` were previously let through to the build dispatch), and additionally dispatches `upload-nexus.yml` so stable tags are published automatically once #1063 lands — the upload dispatch is skipped, not fatal, while that workflow doesn't exist on the tagged commit. Pairs with flagos-ai/build-infra#912 on the shared-workflow side. PR description updated accordingly.

shiptux:
Closing: packages are now built and published centrally from flagos-ai/build-infra, so this repository no longer needs its own publishing step. Thanks for the reviews.

Issue #798, opened 2026-07-15
[BUILD] Packaging follow-ups: Ubuntu 22.04 compat, rpath audit, per-branch backends, vendor runtime deps


shiptux:
Follow-up to #794/#796. Four concerns were raised about the deb/rpm packaging; each was verified against main and the built artifacts. Status and actions below. (I'll take this — please assign to @shiptux.)

### 1. Wheels built on Ubuntu 24.04 are unusable on 22.04 — confirmed
`libtriton.so` requires `GLIBC_2.38` / `GLIBCXX_3.4.32` (verified with `strings`); Ubuntu 22.04 ships glibc 2.35 / GLIBCXX 3.4.30. The deb's `${shlibs:Depends}` does work — the built package declares `libc6 (>= 2.38), libstdc++6 (>= 13.1)`, so 22.04 fails at install time, not at runtime.
- [x] Add an `ubuntu:22.04` matrix variant (#1066) (`DEB_BASE_IMAGE` is already parameterized; the same-base two-stage build handles the py3.10 ABI automatically)
- [ ] Longer term: evaluate a manylinux-style wheel to cover multiple distros with one build

### 2. rpath / hardcoded paths — not reproduced on main (nvidia)
All shipped `.so` files have **no RPATH/RUNPATH**, `DT_NEEDED` lists only system libs, and `patchelf` has zero occurrences on main (repo grep + GitHub code search). The reported `patch_elf` hardcoding lives on vendor-backend branches, so this becomes relevant when packaging those backends.
- [x] Re-verify on a current 0.6.0 wheel (adds libproton.so and instrumentation libs) — done 2026-08-30, see comment: no RPATH/RUNPATH, system-only DT_NEEDED
- [x] Add an `import triton` smoke test to the assembler stages (in Dockerfile.deb since #794; both #1066 rows run it) — CI is currently blind to the whole "installs fine, fails at import" class (covers both #1 and #2)
- [ ] When packaging a vendor backend: audit its `.so` files for RPATH/hardcoded paths; patch to `$ORIGIN`-relative where needed

### 3. Backends live on separate branches — confirmed
Upstream has many per-backend branches (`ascend_*`, `backend/enflame/*`, `feat/mthreads-*`, ...). Packaging files exist only on main, so "add a matrix entry" is not enough for other backends.
[...]

PR #1066, opened 2026-08-28
[BUILD] Add an ubuntu22.04/python3.12 DEB matrix row (FlagOS DEB ABI matrix)

shiptux:
## Why

`build-deb.yml` only builds on `ubuntu:24.04`; the resulting `libtriton.so` requires `GLIBC_2.38` / `GLIBCXX_3.4.32`, so the package declares `libc6 (>= 2.38)` and cannot be installed on Ubuntu 22.04 at all (#798, item 1). The FlagOS DEB ABI matrix (FEP-0019, flagos-ai/community#59) is **ubuntu22.04 + python3.12**, which is also what libtriton_jit#61 and FlagFFT#12 build on — a 22.04 build installs on both 22.04 and 24.04, the reverse does not hold.

## What

- **New matrix row** `ubuntu2204` (`ubuntu:22.04`, python 3.12 from the deadsnakes PPA), artifact `flagtree-nvidia-ubuntu2204-packages`. The original `ubuntu2404` row and its artifact name are unchanged, so upload-nexus (#1063) keeps working.
- **`Dockerfile.deb`** gains `PYTHON_VERSION` / `MAX_JOBS` / `DEB_VERSION_SUFFIX` build args. The interpreter is installed in *both* stages, so the wheel, the .deb and the existing `import triton` smoke test agree on the CPython the package is for.
- **Version suffix**: with `DEB_VERSION_SUFFIX=auto` (default) the package version becomes `0.6.0-1+ubuntu22.04` / `+ubuntu24.04` (Docker CE does the same with `~ubuntu.22.04~jammy`; `+` is used here so the suffixed package sorts above and upgrades an unsuffixed one), so rows built for different releases can coexist in one apt suite instead of colliding on the same filename (flagos-ai/build-infra#600). Set it to "" to keep the plain changelog version.
- **`Depends: python3.X`** derived from the wheel's `cpXY` tag in `debian/rules` instead of the unversioned `python3:any` — the extension is CPython-minor specific, and `python3.12` exists on 22.04 (deadsnakes) and 24.04 natively.
- `build-flagtree.sh` passes the knobs through as env vars (`DEB_BASE_IMAGE`, `PYTHON_VERSION`, `MAX_JOBS`, `DEB_VERSION_SUFFIX`).

## Notes

- The 22.04 row builds the full MLIR wheel like the 24.04 one (~36 min / >4 GB on a 4-core runner); the job-level warning about `ubuntu-latest` headroom applies to both rows. `fail-fast: false` so one row cannot cancel the other.
- The other #798 items (rpath re-check on 0.6.x, backend-branch packaging, vendor runtime deps) are not touched here.

Refs #798.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -50,6 +50,26 @@ jobs:
     # `runs-on: h20`) once one is wired up.
     runs-on: ubuntu-latest
     timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Original row; artifact name unchanged so downstream collectors
+          # (upload-nexus) keep working.
+          - target: ubuntu2404
+            base_image: ubuntu:24.04
+            python_version: ""
+            artifact: flagtree-nvidia-packages
+          # FlagOS DEB ABI-matrix row (FEP-0019): ubuntu22.04 + python3.12.
+          # A 22.04 build also installs on 24.04; the reverse does not hold
+          # (GLIBC_2.38 / GLIBCXX_3.4.32), see #798.
+          - target: ubuntu2204
+            base_image: ubuntu:22.04
+            python_version: "3.12"
+            artifact: flagtree-nvidia-ubuntu2204-packages
+    env:
+      DEB_BASE_IMAGE: ${{ matrix.base_image }}
+      PYTHON_VERSION: ${{ matrix.python_version }}
     steps:
       # The wheel build peaks near the runner's free-disk limit.
       # Ephemeral GitHub-hosted VMs only — never a self-hosted machine.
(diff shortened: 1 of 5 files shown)


shiptux:
Local verification of the new row (`DEB_BASE_IMAGE=ubuntu:22.04 PYTHON_VERSION=3.12 MAX_JOBS=4`, head `40d5853`):

- builds `python3-flagtree-nvidia_0.6.0-1+ubuntu22.04_amd64.deb` (240 MB); `debian/rules` derives `Depends: python3.12` from the `cp312` wheel tag; the in-image `import triton` smoke passes
- resulting Depends: `python3.12, libc6 (>= 2.34), libgcc-s1 (>= 4.2), libstdc++6 (>= 12), zlib1g, python3-filelock, python3-setuptools` — versus `libc6 (>= 2.38), libstdc++6 (>= 13.1)` for the 24.04 build (#798)
- the same .deb installs and `import triton` (3.6.0) works in clean containers on **Ubuntu 22.04** (python3.12 from deadsnakes) and **Ubuntu 24.04**

Two things the first CI run caught and that are now fixed in the branch: `dch` refuses a `~` suffix as a lower version (switched to `+ubuntu<ver>`, which also upgrades an unsuffixed package cleanly, plus `-b`), and on 22.04 the distro pip imports `distutils` under Python 3.12 (the deadsnakes path now gets its own pip/setuptools in /usr/local). The MLIR wheel itself built fine on `ubuntu-latest` for both rows (~35 min each).

zhzhcookie (approved):
LG

PR #1074, opened 2026-08-31
[QC][MTHREADS] SQMMA config selection: prefer a pure M-split on ties

monellz:
When several `(instShape, warpsPerCTA)` candidates tie on instruction count, volume and repetition, the SQMMA config selector picked whichever came first — an N-split (`[4,2]`) for common attention shapes. Kernels that reduce the dot result along axis=1 then pay a cross-warp reduction (shared-memory round trip plus a barrier per tile), while a pure M-split
keeps every output row inside one squad.

Break full ties toward `warpsN == 1`, guarded by `instM >= 32` so the instruction M never degrades to the 16-row forms that measure slower. Everything ahead of the tie (instruction count, volume, repM/repN) is unchanged, so non-tied selections are unaffected. Applied to both selection loops (`tritonmusa-accelerate-matmul` and the TLE `LowerSqmma` copy).

# reproduce script

```python
"""Repro: SQMMA warp-split selection for a reduce-along-axis-1 consumer.

A 64x64x64 fp16 dot at num_warps=8 admits both warpsPerCTA=[8,1] and
[4,2] at identical instruction count, volume and repetition. The kernel
reduces the dot result along axis=1; with [4,2] that reduction crosses
warps through shared memory, with [8,1] every output row stays inside
one squad.

On upstream main: the first candidate wins the tie -> [4,2].
With this patch: full ties break toward warpsN == 1 -> [8,1].
The script prints the selected split from the TTGIR and checks the
numeric result either way.
"""
import re
import torch
import triton
import triton.language as tl

M, N, K = 64, 64, 64


@triton.jit
def _kernel(a_ptr, b_ptr, out_ptr, M: tl.constexpr, N: tl.constexpr,
            K: tl.constexpr):
    offs_m = tl.arange(0, M)
    offs_n = tl.arange(0, N)
    offs_k = tl.arange(0, K)
    a = tl.load(a_ptr + offs_m[:, None] * K + offs_k[None, :])
    b = tl.load(b_ptr + offs_k[:, None] * N + offs_n[None, :])
    sa = tle.gpu.alloc((M, K), dtype=tl.float16, layout=None, init_value=a)
    sb = tle.gpu.alloc((K, N), dtype=tl.float16, layout=None, init_value=b)
    acc = tle.gpu.wgmma(sa, sb, tl.zeros((M, N), dtype=tl.float32))
    row = tl.sum(acc, axis=1)
    tl.store(out_ptr + offs_m, row)


def main():
    torch.manual_seed(0)
    a = torch.randn(M, K, dtype=torch.float16, device="musa")
    b = torch.randn(K, N, dtype=torch.float16, device="musa")
    out = torch.empty(M, dtype=torch.float32, device="musa")
    h = _kernel[(1, )](a, b, out, M, N, K, num_warps=8)
    torch.musa.synchronize()
    ref = (a.float() @ b.float()).sum(dim=1)
    ok = torch.allclose(out, ref, atol=1e-1, rtol=1e-2)

    ttgir = h.asm["ttgir"]
    sqmma = sorted(set(
        re.findall(r"sqmma[^>\n]*warpsPerCTA\s*=\s*\[(\d+),\s*(\d+)\]",
                   ttgir, re.IGNORECASE)))
    if not sqmma:
        sqmma = sorted(set(
            re.findall(r"warpsPerCTA\s*=\s*\[(\d+),\s*(\d+)\]", ttgir)))
    print(f"sqmma-msplit repro: {'PASS' if ok else 'FAIL'} "
          f"(sqmma warpsPerCTA: {sqmma}; ('4','2') without this patch, "
          f"('8','1') with it)")
    assert ok


if __name__ == "__main__":
    import torch_musa  # noqa
    main()

```


sunnycase (approved):
LGTM

PR #1075, opened 2026-08-31
[QC][MTHREADS] Lower PH1 swizzled shared access through LinearLayout

monellz:
The PH1 swizzled shared-memory path lowered every element through a runtime globalize/linearize/swizzle chain (~8 div/rem/xor ops per element, with the div/rem strength-reduced to shifts). Constant folding cannot rescue it: the address mixes a runtime base (shared-memory base, plus the rotating slot offset under multibuffering) with a compile-time element offset, and `%`, `/`, `^` do not distribute over that sum. This PR makes the identity hold
first, then emits the folded form:

- **AllocateSharedMemory**: raise PH1 swizzled allocs' alignment from the 256B line to the full swizzle window (e.g. SG_16B → 4KB). With base `B` aligned to window `W` and offset `o < W`, `swz(B+o) = B + swz(o)` — the swizzle becomes base-independent and evaluable on buffer-relative coordinates, which are compile-time constants per element.
- **LoadStoreOpToLLVM**: a LinearLayout fast path composes the register layout with the physical shared layout so each address is `threadBase ^ constant`. Static subviews fold entirely; a runtime view offset (rotating multibuffer slot) costs one linearize+swizzle chain per access site, conservatively widened so the XOR composition stays valid. The async-copy swizzled branch gets the same treatment (each 16B word = `base ^ immediate`). The scalar chain remains as the fallback for shapes the fast path cannot prove.
- **Two view-vs-alloc fixes** the fast path relies on: the shared LinearLayout and the TME swizzle config must be resolved against the *alloc* shape, not a subslice view's shape — a `[64,128]` column view of a `[64,256]` bf16 buffer otherwise resolves a self-consistent but wrong swizzle-group config (SG_32B vs the SG_16B the SQMMA hardware uses), scrambling tile stores through the view. Register layouts are also extended to the physical shape explicitly instead of trusting `invertAndCompose`'s truncation.

# reproduce script

```python
"""Repro: PH1 swizzled shared-memory access lowering cost.

The kernel fills both SQMMA operands from register tensors
(init_value), which lowers as a tensor store into a PH1-swizzled shared
buffer, then multiplies them with wgmma and checks against torch.

On upstream main: correct, but every access goes through the runtime
swizzle chain -- for this small kernel the LLIR contains ~64 xors of
two runtime values (plus the div/mod chains strength-reduced to
shifts).
With this patch: identical results; the stores lower through the
LinearLayout fast path where each address is threadBase ^ constant, and
the runtime-operand xor count collapses to ~2.
"""
import re
import torch
import triton
import triton.language as tl

M, N, K = 64, 64, 64


@triton.jit
def _kernel(a_ptr, b_ptr, c_ptr, M: tl.constexpr, N: tl.constexpr,
            K: tl.constexpr):
    offs_m = tl.arange(0, M)
    offs_n = tl.arange(0, N)
    offs_k = tl.arange(0, K)
    a = tl.load(a_ptr + offs_m[:, None] * K + offs_k[None, :])
    b = tl.load(b_ptr + offs_k[:, None] * N + offs_n[None, :])
    # init_value stores a register tensor into a PH1-swizzled shared buffer
    sa = tle.gpu.alloc((M, K), dtype=tl.float16, layout=None, init_value=a)
    sb = tle.gpu.alloc((K, N), dtype=tl.float16, layout=None, init_value=b)
    acc = tle.gpu.wgmma(sa, sb, tl.zeros((M, N), dtype=tl.float32))
    tl.store(c_ptr + offs_m[:, None] * N + offs_n[None, :], acc)


def main():
    torch.manual_seed(0)
    a = torch.randn(M, K, dtype=torch.float16, device="musa")
    b = torch.randn(K, N, dtype=torch.float16, device="musa")
    c = torch.empty(M, N, dtype=torch.float32, device="musa")
    h = _kernel[(1, )](a, b, c, M, N, K, num_warps=4)
    torch.musa.synchronize()
    ref = a.float() @ b.float()
    ok = torch.allclose(c, ref, atol=1e-1, rtol=1e-2)

    llir = h.asm["llir"]
    const_xor = len(re.findall(r"xor i32 %\S+, \d", llir))
    run_xor = len(re.findall(r"xor i32 %\S+, %", llir))
    print(f"ph1-swizzle repro: {'PASS' if ok else 'FAIL'} "
          f"(llir runtime-operand xors={run_xor}, constant-operand "
          f"xors={const_xor}; ~64 runtime xors without this patch, "
          f"~2 with it)")
    assert ok


if __name__ == "__main__":
    import torch_musa  # noqa
    main()

```


sunnycase (approved):
LGTM

PR #1076, opened 2026-08-31
[BUILD] deb: strip DWARF from libtriton/libproton and skip dh_dwz

shiptux:
## Why

The deb ships `libtriton` **724 MB — 537 MB of which is `.debug_*` sections** — and `libproton` completely unstripped, because `debian/rules` excludes them from `dh_strip` (Triton's JIT needs their symbol tables). `dh_dwz` also leaves its multifile under `usr/lib/debug/.dwz/` **inside the runtime package**. Both were found while re-auditing rpath/hardcoded paths for #798 (the absolute `/src/...`, `/root/.triton/llvm/...` strings in the binaries are exactly this DWARF data).

## What

- After `dh_strip`, run `strip --strip-debug` on the two excluded objects: it removes only the DWARF sections and **keeps `.symtab`/`.dynsym`**, which is what the JIT lookups use.
- No-op `override_dh_dwz` (nothing left for it to compress, and its output does not belong in the runtime package).
- `lintian-overrides` documenting the deliberately kept symbol tables (lintian counts a present `.symtab` as "unstripped").

## Verified

Built with the cached wheel (validated on top of #1066's branch; the change applies to `main` independently):

| | before | after |
|---|---|---|
| `libtriton…so` | 724 MB | **187 MB** |
| `.deb` | 240 MB | **80 MB** |
| lintian `debug-suffix-not-dbg` | 1 | 0 |

The in-image smoke (`apt-get install` + `import triton`) still passes, and a separately installed package imports `triton`, `triton.compiler` and the nvidia driver after `--strip-debug`.

No conflict with #1066 (different hunks of `debian/rules`); whichever merges second rebases trivially.

Refs #798.

### DIFF
--- a/packaging/debian/python3-flagtree-nvidia.lintian-overrides
+++ b/packaging/debian/python3-flagtree-nvidia.lintian-overrides
@@ -1,0 +1,6 @@
+# libtriton/libproton keep their symbol tables on purpose: Triton's JIT does
+# runtime symbol lookups in them. Only the DWARF sections are removed
+# (strip --strip-debug in debian/rules); lintian counts a present .symtab as
+# "unstripped".
+unstripped-binary-or-object [usr/lib/python3/dist-packages/triton/_C/libtriton.cpython-312-x86_64-linux-gnu.so]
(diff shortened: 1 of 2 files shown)


zhzhcookie (approved):
LG

PR #1082, opened 2026-09-01
[CICD] Bump packaging baseline to 0.6.1 and plumb the wheel version

shiptux:
## What

- Bump the RPM spec and Debian changelog from 0.6.0 to the 0.6.1 release.
- Derive the wheel version from the packaging files and pass it into the docker wheel builds as `FLAGTREE_WHEEL_VERSION` (RPM: from the spec's `Version:`; deb: from the changelog's upstream version).

## Why

`setup.py` falls back to a hardcoded `0.6.0` whenever `FLAGTREE_WHEEL_VERSION` is unset, and the deb/rpm pipelines never set it — so they produce 0.6.0 wheels regardless of the checked-out tag. This is the mechanism behind the packaging baseline lagging the release tags: even a build on the `0.6.1` tag ships a wheel that claims 0.6.0.

With this change the packaging files are the single source of truth: bumping the spec/changelog version is all a future release refresh needs, and the wheel inside the package always matches the package version.

## Validation

Local build from the `0.6.1` tag source with this packaging, `RPM_BASE_IMAGE=openeuler/openeuler:24.03-lts`:

- wheel built as `flagtree-0.6.1-cp311-cp311-linux_x86_64.whl` (previously would have been 0.6.0);
- output `python3-flagtree-nvidia-0.6.1-1.oe2403.x86_64.rpm` + src.rpm;
- the Dockerfile's built-in install + `import triton` smoke stage passed.

Empty `FLAGTREE_WHEEL_VERSION` keeps the previous fallback behavior, so builds invoked without the helper scripts are unaffected.


zhzhcookie (dismissed):
LG

shiptux:
Rebased onto current main. The conflict came from #1066 and #1076 landing in the meantime — both touch `build-flagtree.sh`. Resolved by keeping both sides: the build-environment knobs from #1066 and the changelog-derived `WHEEL_VERSION` from this PR, with the two merged into one build line. Verified the version extraction still yields the changelog's upstream version and the script parses clean.

zhzhcookie (approved):
LG

PR #1104, opened 2026-09-06
[TLE][MTHREADS] Support multi-fields pipe and multi-roles ws

### LINKED ISSUE #1028: [mthreads]tle.pipe() does not support multiple fields; tle.gpu.warp_specialize() does not support muti-role specialization
### Environment
- **OS:** Ubuntu 22.04.5 LTS
- **Python:** 3.10.12
- **FlagTree:** 0.6.0+mthreads.gite9525632
- **Triton:** 3.6.0
- **PyTorch:** 2.7.1
- **torch_musa:** 2.7.1
- **MUSA SDK:** 5.1.0
- **Device:** 8 × Moore Threads MTT S5000
## Description
The same API is used by NVIDIA-style TLE pipelines where a pipe carries multiple shared-memory buffers simultaneously (for example, separate buffers for different pipeline stages or operands).
On Mthreads, the API is available, but creating a pipe with multiple fields fails during compilation, and it raises "initial mthreads tle.pipe requires exactly one payload field".
## Minimal reproduction
This creates one pipeline object containing multiple shared-memory objects (a and b) that can be accessed together through pipeline operations.The same API fails on Mthreads during compilation instead of creating a valid multi-field pipe.It appears that the API exists, but multi-field pipes are not currently supported.
```
import torch
import triton
import triton.language as tl


@triton.jit
def kernel(x, y):
    a = tle.gpu.alloc(
        [2, 128],
        dtype=tl.float32,
        scope=tle.gpu.smem,
    )
    b = tle.gpu.alloc(
        [2, 128],
        dtype=tl.float32,
        scope=tle.gpu.smem,
    )

    p = tle.pipe(
        capacity=4,
        scope="cta",
        name="multi",
        a=a,
        b=b,
    )


x = torch.empty(1, device="musa")
y = torch.empty(1, device="musa")

kernel[(1,)](x, y, num_warps=4)
```

## Description
While porting NVIDIA MegaMoE kernels to MUSA, I found that tle.gpu.warp_specialize() does not support three specialized roles.NVIDIA MegaMoE uses three independent warp roles within one CTA:
default role → Loader
partition 0 → Dispatch
partition 1 → Math
This allows different warp groups to execute different tasks concurrently inside one CTA.
On Mthreads, warp_specialize() is available, but the same three-role configuration fails, and it raises "mthreads TLE static warp_specialize requires exactly one producer partition".
## Minimal reproduction
```
import torch
import triton
import triton.language as tl


@triton.jit
def consumer(x):
    v = tl.load(x)
    tl.store(x, v)

@triton.jit
def worker0(x):
    pass

@triton.jit
def worker1(x):
    a = tl.full((), 1.0, tl.float32)
    b = a + 2.0

@triton.jit
def kernel(x):
    tle.gpu.warp_specialize(
        [
            (consumer, (x,)),
            (worker0, (x,)),
            (worker1, (x,)),
        ],
        [2, 2],
        [64, 64],
    )


x = torch.zeros((1,), device="musa")
kernel[(1,)](x)
torch.musa.synchronize()
print(x)
```

### LINKED ISSUE #1109: [BUG][mthreads] Block-pointer load+store with integer `//`/`%` zeroes the last valid element of tail blocks
## Description

When a kernel uses `tl.make_block_ptr` for **both** the loads and the store (with `boundary_check`), and the computation is an **integer** division family op (`//` or `%`, int32/int64), and the number of elements is **not divisible** by the block size, the **last valid element of the tail block is always written as 0**.

The result value at that position is not garbage — it is deterministically `0`, which suggests the last valid lane is being treated as out-of-bounds (padded to 0 on load) somewhere in the block-pointer lowering pipeline, but only when the integer-division expansion is present.

## Minimal reproduction

```python
import torch
import torch_musa  # noqa: F401

import triton
import triton.language as tl


@triton.jit
def div_kernel(x_ptr, y_ptr, out_ptr, n, BLOCK: tl.constexpr):
    off = tl.program_id(0) * BLOCK
    xb = tl.make_block_ptr(x_ptr, (n,), (1,), (off,), (BLOCK,), (0,))
    yb = tl.make_block_ptr(y_ptr, (n,), (1,), (off,), (BLOCK,), (0,))
    ob = tl.make_block_ptr(out_ptr, (n,), (1,), (off,), (BLOCK,), (0,))
    x = tl.load(xb, boundary_check=(0,))
    y = tl.load(yb, boundary_check=(0,))
    tl.store(ob, x // y, boundary_check=(0,))


n, BLOCK = 100, 64  # 2 blocks; tail block has 36 valid elements
torch.manual_seed(0)
x = torch.randint(1, 1000, (n,), dtype=torch.int32)
y = torch.randint(1, 10, (n,), dtype=torch.int32)
ref = torch.div(x, y, rounding_mode="floor")

xd, yd = x.to("musa"), y.to("musa")
out = torch.empty_like(xd)
div_kernel[(triton.cdiv(n, BLOCK),)](xd, yd, out, n, BLOCK=BLOCK)
got = out.cpu()

print(f"out[{n-1}] = {got[-1].item()}   expected = {ref[-1].item()}")
bad = (got != ref).nonzero().flatten().tolist()
assert not bad, f"mismatch at indices {bad}"
print("PASS")
```

Observed output:

```
out[99] = 0   expected = 112
AssertionError: mismatch at indices [99]
```

Exactly one element is wrong — the last valid element of the tail block — and it is always `0`.

## What does NOT trigger it (controls)

All of the following are correct on the **same data and launch config** (n=100, BLOCK=64, num_warps=4):

| variant | result |
|---|---|
| block-ptr load + block-ptr store, int32 `//` | **WRONG** (`out[99] = 0`) |
| pointer+mask load + pointer+mask store, int32 `//` | correct |
| block-ptr load + pointer+mask store, int32 `//` | correct |
| pointer+mask load + block-ptr store, int32 `//` | correct |
| block-ptr load + block-ptr store, **fp32 `/`** | correct (max rel err ~1e-7) |

So it requires the full block-pointer path on both sides *and* integer division; replacing either side with the pointer+mask style, or switching to float division, makes the bug disappear. This points at a pass-level interaction between the block-pointer lowering chain and the integer-division expansion rather than a plain boundary-check off-by-one.

## Whether it triggers depends on shape / num_warps (looks flaky under autotuning)

Sweep over `{//, %} × {int32, int64} × (n, BLOCK) × num_warps ∈ {1, 4, 8}`: 85 of 120 1-D cases fail, always with the same signature (last valid tail element = 0). Examples (int32 `//`; int64 and `%` behave identically):

| n | BLOCK | w=1 | w=4 | w=8 |
|---|---|---|---|---|
| 1000 | 128 | WRONG | ok | WRONG |
| 100 | 64 | WRONG | WRONG | WRONG |
| 1025 | 1024 | WRONG | WRONG | WRONG |
| 2049 | 256 | WRONG | WRONG | ok |
| 33 | 32 | ok | WRONG | WRONG |

Because triggering depends on `(n, BLOCK, num_warps)`, integer `div`/`remainder`/`floor_divide` tests in FlagGems (which autotune over block sizes/warps) appear **flaky**, while the underlying miscompile is fully deterministic for a fixed config.

Note when sweeping with `%`: if the reference value of the last tail element happens to be `0`, the case falsely "passes" (the bug writes `0` there) — worth keeping in mind for any regression test.

## 2-D block pointers are affected more severely

With 2-D `make_block_ptr` (`boundary_check=(0, 1)`) and int32 `//`, entire boundary regions are zeroed, not just one element:

| M | N | BM | BN | num_warps | wrong elements |
|---|---|---|---|---|---|
| 13 | 100 | 16 | 64 | 4 | 112 |
| 100 | 33 | 32 | 32 | 4 | 132 |
| 7 | 1000 | 8 | 256 | 8 | 1001 |

All wrong elements read back as `0`.

## Environment

- FlagTree: `main` @ `7a8d56c33`
- GPU: MTT S5000 (arch `ph1`, capability (3,1))
- torch 2.7.1 / torch_musa 2.7.1+5ee0a64, Python 3.10

Kylin1207:
- Closes #1028
- Closes #1109


zhzhcookie (approved):
LGTM

sunnycase (approved):
LGTM

PR #1115, opened 2026-09-07
[CI][Nvidia] Optimize concurrency grouping for nvidia3.6 workflow

liuhaoyutz:
Refine the concurrency group logic to better handle different event types:
- pull_request: Group by PR number to cancel outdated runs from the same PR.
- push: Group by commit SHA to allow parallel runs from different merges.
- schedule/others: Fall back to run_id to keep each scheduled run independent.


zhzhcookie (approved):
LGTM

PR #703, opened 2026-06-17
[BACKEND][TileIR] Add opt-in TileIR backend integration

KingsleyLiu-NV:
## Summary

This PR integrates [Triton-to-Tile-IR](https://github.com/triton-lang/Triton-to-Tile-IR/tree/main) into FlagTree as an independent `tileir` backend.

TileIR has its own compiler and driver and is installed alongside the existing NVIDIA and AMD backends. The common `python/triton` layer only provides backend-neutral routing, compiler, driver, and language-extension hooks. TileIR-specific policy and implementation remain under `third_party/tileir`.

Runtime behavior is unchanged unless `FLAGTREE_USE_TILEIR=1` is set.

The default source installation includes `nvidia`, `amd`, and `tileir`.
`FLAGTREE_BACKEND=tileir` is also supported as an explicit build mode; it keeps
the NVIDIA and AMD backends installed because TileIR uses per-kernel fallback
to the native NVIDIA path.

## Design

TileIR is installed as an independent backend with its own `TileIRBackend` and `TileIRDriver`. On NVIDIA systems, `CudaDriver` remains the active hardware driver and produces the initial `cuda` target. Routing may select a `tileir` target for an individual kernel without replacing or modifying the NVIDIA backend.

The shared Python changes are backend-neutral hooks:

```text
python/triton/runtime/jit.py
├── python/triton/backends/__init__.py::route_target()
│   └── TileIRBackend.route_target()
│       └── third_party/tileir/backend/router.py
└── python/triton/compiler/compiler.py
    ├── get_backend(final_target)
    ├── backend.make_ir(...)
    └── get_driver(final_target, active_driver)
```

The common interfaces under [`python/triton/backends`](https://github.com/KingsleyLiu-NV/FlagTree/tree/feature/flagtree-tileir-integration/python/triton/backends) do not import TileIR implementation code. TileIR-specific routing, compilation, and driver behavior remain under `third_party/tileir`.

With `FLAGTREE_USE_TILEIR=1`:

- CUDA kernels without TLE route to TileIR.
- Kernels using only the supported top-level `tle` view/token subset route to TileIR.
- Other or unknown TLE usage remains on native NVIDIA.
- Non-CUDA targets remain unchanged.

The policy is implemented in the [TileIR backend router](https://github.com/KingsleyLiu-NV/FlagTree/blob/feature/flagtree-tileir-integration/third_party/tileir/backend/router.py).

[...]

PR #261, opened 2025-12-31
[TLE] Add initial tle design

sunnycase:
This pull request introduces several enhancements and new features to the Triton and TritonGPU dialects, focusing on extensibility, shared memory handling, and infrastructure for new plugins and operations. The most significant changes are the addition of the "tle" plugin, support for custom load hints (notably for shared memory), a new `tma_copy` operation, and a new pass for processing shared memory hints. Below are the key changes grouped by theme:

**Plugin and Build System Enhancements:**
- Added the "tle" plugin to the build system by updating `CMakeLists.txt` to include and build the new plugin under `third_party/tle`. This ensures the new plugin is available both for Python module and non-Python module builds. [[1]](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR286-R289) [[2]](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR480-R481)

**Triton IR and Operation Extensions:**
- Extended `TT_LoadOp` in `TritonOps.td` to support a new `flagtree_hints` string attribute, with corresponding builder overloads in `Ops.cpp`. This allows passing special hints (such as `"shared_memory"`) to load operations, enabling downstream optimizations and transformations. [[1]](diffhunk://#diff-4b81ba7a5ae52409ed87dd62d285d8758186d2a3af9601767a891e2dab45f558L250-R276) [[2]](diffhunk://#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930eR48-R56) [[3]](diffhunk://#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930eR65-R74) [[4]](diffhunk://#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930eR83-R92) [[5]](diffhunk://#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930eR101-R111) [[6]](diffhunk://#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930eR125-R139)
- Updated the rewrite and pattern-matching infrastructure to propagate and handle the new `flagtree_hints` attribute in load operations, ensuring correct transformation and codegen.

**TritonGPU Dialect and Passes:**
[...]

PR #1116, opened 2026-09-07
[PPU][MTHREADS] Update low-precision float (fp8/fp4) support via backend capability contract

zeroherolin:
### Summary

This PR puts low-precision float (fp8/fp4) support on the PPU and MThreads backends under an explicit capability contract, and extends it where gaps remained. Each backend now declares the casts and dot combinations it actually supports; the semantic layer rejects undeclared combinations with the backend's diagnostic and warns at compile time whenever a non-native path runs — nothing degrades silently anymore. Pre-existing paths (PPU's FP16-promotion fp8 dot, MThreads' SQMMA/WMMA fp8 lowering with FMA fallback, fp8 casts, fp4 upcast) keep their behavior, now declared and fixed where broken; newly added are fp8e4nv on PPU cap80–88 via a bit-exact software cast, a working FP4→fp16 upcast below cap89, a rewritten interpreter fp-conversion baseline, runnable fp8/fp4 tutorials, and a fallback knob that restores the upstream rules.

### Changes

- **Common**
  - *new*: DotCap / DotSupport tri-state (NATIVE / NON_NATIVE / UNSUPPORTED) with optional resolve_dot / resolve_dot_scaled codegen queries — UNSUPPORTED combinations raise with the backend's diagnostic, NON_NATIVE ones emit a compile-time warning (native=false); backends that do not register the callbacks keep the upstream rules.
  - *new*: capability declaration fields supported_fp8_cast_dtypes / custom_cast_fp8_dtypes (cast whitelist and software-cast routing) and async_copy_dtypes / descriptor_dtypes (data movement; declaration-only at this stage), all shipped with h.metadata.
  - *new*: semantic gating — tl.dot queries resolve_dot and tl.dot_scaled queries resolve_dot_scaled; fp8 casts route by the declared whitelist instead of the hardcoded fp8e4b15-only special case.
  - *new*: interpreter numerics baseline — _convert_float rewritten with exact fp64 decode then RTNE/RTZ encode and per-format fp8 special-value handling; the original implementation is kept as _convert_float_legacy (fixes the first three items under Bug fixes).
  - *new*: fallback knob FLAGTREE_LOW_PRECISION_FLOAT (default 1) — 0 restores the upstream interpreter routine, PPU/MThreads stop registering the resolvers and PPU stops declaring the cap80–88 software fp8e4nv cast; added to CACHE_INVALIDATING_ENV_VARS.
  - *new*: runnable tutorials python/tutorials/mixed_precision/01-fp8.py / 02-fp4.py covering capability queries, saturating casts, fp8 matmul, FP4 decode and dot_scaled.
- **PPU 810E**
  - *new*: fp8e4nv opened on cap80–88 through a bit-exact software cast (convert_custom_types routing); native cvt from cap89 unchanged — the type was previously rejected outright below cap89.
  - *unified*: resolve_dot / resolve_dot_scaled — int8 NATIVE; same-type fp16 / bf16 / fp32 / fp64 NATIVE; fp8 NATIVE from cap89, NON_NATIVE below (FP16-promotion path, warns); dot_scaled declared NATIVE for cap89 e2m1 combinations, NON_NATIVE decompose otherwise (narrowed to both-sides e2m1 in #1189); every other combination UNSUPPORTED with the product name and native alternatives.
- **MThreads S5000**
  - *unified*: fp8 dual whitelist — supported_fp8_dtypes keeps gating dot, new supported_fp8_storage_dtypes gates storage; cast and data-movement families derive from the storage whitelist; custom_fp8_dtypes renamed to custom_cast_fp8_dtypes.
  - *unified*: resolve_dot / resolve_dot_scaled — int8 NATIVE; same-type fp16 / bf16 / fp32 / fp64 NATIVE; same-type fp8 with fp32 accumulator on instruction-tile shapes NATIVE, mixed / non-fp32-acc / shape-miss NON_NATIVE (FMA fallback, warns with reason); dot_scaled always NON_NATIVE decompose.
  - *unified*: enable_fp8_burst2 takes effect through options — the field is plumbed to the C++ lowering as a pass option (the pre-existing env-var-only behavior under Bug fixes).
  - *unified*: spec overlays (compiler / semantic / interpreter) synced with the common layer; GetEnv.hpp mirrors the fallback knob.

### Bug fixes

- Common (upstream Triton unless noted)
  - Interpreter fp8 rounding (still present in v3.6.0): float→fp8 rounded half-up instead of RTNE and ignored mantissa carry — exhaustive compare: 7526 wrong values for f16→e4nv, 5456 for f16→e5m2; the rewrite passes the exhaustive compare.
  - Interpreter fp8 subnormals: 1496 wrong values for f16 subnormals→e5m2; e5m2 codes 0x01–0x03 decoded to 0 or half their value — unified underflow-shift handling.
  - Interpreter fp8 specials: E4M3FN 0x7F/0xFF decoded as ±480 instead of NaN; fnuz zero could be encoded as a NaN code — per-format special-value conventions (ieee / fn / fnuz / none), fnuz zero always encodes +0.
  - Interpreter load signature (FlagTree TLE, #261): semantic load gained flagtree_hints but the interpreter did not follow — tl.load raised TypeError in interpreter mode.
  - Language-extension proxy dunder (FlagTree TileIR, #703): inspect / copy probing __xxx__ attributes hit the backend-module lookup and raised — dunder lookups now raise AttributeError directly.
- **PPU 810E**
  - FP4 decode missing its capability gate: the FP4→fp16 path unconditionally used the cap89-only e4m3 cvt — an illegal instruction on cap80 (ported from upstream NVIDIA lowering, which has no sub-sm89 FP4 case).
  - fp8 conversion table wrongly static: table entries vary with computeCapability, so compiling cap80 then cap89 in one process reused the first table (same flaw exists upstream).
  - ppu-llc error path NameError: log_file was never defined, so the real compile error was masked (introduced in #843); unused import signal removed.
- **MThreads S5000** (all from #577)
  - SQMMA fp8 operand typing: fp8 data loaded as int8 left shared memory typed int8, mismatching the fp8 type the dot consumes — MemDescReinterpretOp inserted and type propagation extended.
  - WMMA accepted fp16 accumulators and generated illegal IR — non-fp32 results (except PH1 f16×f16) now fall back to FMA.
  - WMMA verifier only checked equal bit-width, letting e4m3×e5m2 through — same-type check added.
  - User-provided supported_fp8_dtypes was ignored when deriving the storage whitelist — the user value now takes precedence.
  - enable_fp8_burst2 existed in options but only the TRITON_MUSA_ENABLE_FP8_BURST2 env var was read — the field is now passed via pass option and the env read removed.

### Compatibility

- The contract is optional: backends that do not register it keep the legacy dot/cast rules, and defaults preserve existing behavior for all other backends.
- Interpreter-mode numerics change for every backend as a correctness fix (strict RTNE and per-format fp8 special values); FLAGTREE_LOW_PRECISION_FLOAT=0 restores the previous behavior.

### Validation

- 810E: exhaustive 2²⁴ f32→fp8e4nv software-cast checks with strict RTNE and clamping on cap80, compile-error negatives and warnings, both tutorials.
- S5000 (feature-branch tip, with #1189 applied): full test_core, test_compile_errors, test_conversions, test_mxfp, TLE and runtime suites, tutorials 01/02; same-type fp8 dot confirmed native at the llir level; enable_fp8_burst2 verified at the llir level (1 selects the f162e4m3.rn.bst2 intrinsic, 0 the scalar variant; the env knob feeds the default); FLAGTREE_LOW_PRECISION_FLOAT=0/1 both verified.
- Device-free: test_interpreter_fp_conversion pins the RTNE tie points, NaN codes, satfinite, full 256-code roundtrips for all five fp8 formats, and the fallback-knob dispatch.

Supersedes #1112 / #1113 / #1114.


third_party/ppu/language/ppu/utils.py
```diff
@@ -123,3 +123,100 @@ def convert_custom_float8_internal(arg, dst_ty, fp_downcast_rounding, has_minx2,
...
+            return up.to(dst_sca, _semantic=_semantic)
+        if dst_sca.is_fp8e5():
+            rounding = "rtz" if _rounding_is_rtz(fp_downcast_rounding) else "rtne"
+            return up.to(dst_sca, fp_downcast_rounding=rounding, _semantic=_semantic)
+        raise ValueError(f"cast from fp8e4nv to {dst_sca} is not supported on this product")
+    if dst_sca.is_fp8e4nv():
+        if not src_sca.is_floating():
+            raise ValueError(f"cast from {src_sca} to fp8e4nv is not supported on this product")
+        # f16/bf16 widen exactly (single rounding); an fp64 source rounds twice
+        x = arg if src_sca.is_fp32() else arg.to(core.float32, _semantic=_semantic)
```
tianyi-dialect: ## FP64 → fp8e4nv 软件路径双重舍入

### 问题描述

**源码位置：**`third_party/ppu/language/ppu/utils.py:219-221`。

cap80 的新增软件路径接受 FP64 源类型，但先调用 `arg.to(core.float32)`，默认用 RTNE 转为 FP32，再执行目标 FP8 舍入。决定最终舍入方向的低位已在第一次转换中丢失。

- RTNE：真实输入略大于 FP8 中点，先转 FP32 后恰好落在中点，最终错误选择偶数邻居。
- RTZ：真实输入略小于一个 FP8 可表示值，先转 FP32 后变成该值，最终输出大于正输入，违反向零舍入方向。

### 问题复现

**输入与实测结果：**

| 模式 | FP64 输入         | 正确 FP8 编码及数值 | 当前实际编码及数值 |
| ---- | ----------------- | ------------------- | ------------------ |
| RTNE | `1.0625 + 2**-30` | `0x39`，1.125       | `0x38`，1.0        |
| RTZ  | `1.125 - 2**-30`  | `0x38`，1.0         | `0x39`，1.125      |

测试直接从 GPU 上的 FP64 tensor 加载数据，避免 Python 标量在进入 kernel 前已经变成 FP32。

**完整最小脚本：**[test_fp64_cast.py]。

```python
import pytest
import torch
import triton
import triton.language as tl


@pytest.fixture(autouse=True)
def fresh_cache(tmp_path):
    assert triton.knobs.language.low_precision_float, "set FLAGTREE_LOW_PRECISION_FLOAT=1"
    with triton.knobs.cache.scope():
        triton.knobs.cache.dir = str(tmp_path / "cache")
        yield


@pytest.fixture
def cap80():
    target = triton.runtime.driver.active.get_current_target()
    if target.arch != 80:
        pytest.skip("GPU reproducer requires PPU cap80")


@triton.jit
def cast_kernel(X, Y, N: tl.constexpr, DST: tl.constexpr, ROUND: tl.constexpr):
    i = tl.arange(0, 128)
    x = tl.load(X + i, i < N, other=0.0)
    y = x.to(DST, fp_downcast_rounding=ROUND)
    tl.static_assert(y.dtype == DST, "cast returned the wrong dtype")
    tl.store(Y + i, y.to(tl.uint8, bitcast=True), i < N)


@pytest.mark.parametrize("rounding,value,expected", [
    ("rtne", 1.0625 + 2**-30, 0x39),
    ("rtz", 1.125 - 2**-30, 0x38),
])
def test_fp64_to_e4nv(cap80, rounding, value, expected):
    x = torch.tensor([value], dtype=torch.float64, device="cuda")
    y = torch.empty(1, dtype=torch.uint8, device="cuda")
    cast_kernel[(1,)](x, y, 1, tl.float8e4nv, rounding)
    actual = y.cpu().item()
    assert actual == expected, f"{rounding}: actual=0x{actual:02x}, expected=0x{expected:02x}"
```

**执行命令：**：

```bash
python3 -m pytest -vs test_fp64_cast.py
```

```text
AssertionError: rtne: actual=0x38, expected=0x39
AssertionError: rtz: actual=0x39, expected=0x38
2 failed
```

### 修复建议

实现保留最终舍入语义的 FP64 下转；若暂不支持，应明确拒绝该源类型，不能接受后静默返回错误结果。

third_party/ppu/python/test/unit/language/test_conversions.py
```diff
@@ -276,7 +276,9 @@ def test_typeconvert_upcast(src_dtype, dst_dtype, device):
     # On HIP, fp8e4nv upcasting to fp32 is only supported on CDNA4, and
     # fp8e4nv upcasting to bf16 and fp16 is only supported on CDNA3 and CDNA4.
     if is_cuda() or is_ppu():
-        if ((src_dtype == 'float8e4nv' and torch.cuda.get_device_capability(0) < (8, 9))
+        # On PPU, fp8e4nv casts are software-lowered from cap80
+        e4nv_error_cc = (8, 0) if is_ppu() else (8, 9)
```
tianyi-dialect: 建议当前 PR 中涉及到的 ut 改动，都将 is_cuda 和 is_ppu 的测试执行路径拆分开，避免与 NV 测试路径耦合。例如：
```python
if is_cuda():
     ...
elif is_ppu():
    ...
```

tianyi-dialect:
LGTM

zhzhcookie (approved):
LGTM

PR #1129, opened 2026-09-08
[FlagTune] Preserve caller-stream ordering in CUDA graph benchmarks

### LINKED ISSUE #1127: [Bug] do_bench_cudagraph starts warmup without waiting for the caller stream
## Problem

The public [do_bench_cudagraph implementation](https://github.com/flagos-ai/FlagTree/blob/main/python/triton/testing.py) switches to a newly created stream and immediately invokes the warmup:

```python
with torch.cuda.stream(torch.cuda.Stream()):
    # warmup
    fn()
```

There is no dependency on work already queued on the caller stream before this first call. If the benchmark reads inputs produced by that work, or writes storage that an earlier operation still reads, the warmup can race with the caller. Synchronization later in the helper cannot repair a result already corrupted by that race.

Please establish a caller-to-benchmark stream dependency before warmup. The relevant stream creation pattern was verified in the repository's public main-branch source when filing this report.

## Minimal reproduction

The following synthetic example requires only PyTorch and the benchmark helper, with no model or application code. It deliberately queues an earlier read and makes the benchmark write the same buffer. The earlier read should retain the original value.

```python
import torch
from triton.testing import do_bench_cudagraph

# Initialize the helper before the ordering experiment.
scratch = torch.zeros(1024, device="cuda")
do_bench_cudagraph(lambda: scratch.zero_(), rep=1, n_retries=1)

failures = []
for use_default in (False, True):
    for attempt in range(4):
        buf = torch.ones(1024, device="cuda")
        torch.cuda.synchronize()
        caller = (
            torch.cuda.default_stream()
            if use_default
            else torch.cuda.Stream()
        )
        with torch.cuda.stream(caller):
            torch.cuda._sleep(100000000)
            prior_read = buf.clone()
            do_bench_cudagraph(lambda: buf.fill_(2), rep=1, n_retries=1)
        torch.cuda.synchronize()
        actual = prior_read.unique().tolist()
        if actual != [1.0]:
            failures.append((use_default, attempt, actual))

assert not failures, failures
```

Expected: every earlier read contains only `1.0`. A read containing `2.0` exposes the missing ordering.

This is a timing-dependent race. The private `torch.cuda._sleep` API is used only to widen the reproduction window; its availability and timing may vary by backend. A clean trial alone does not prove the streams are correctly ordered.

## Proposed fix

Replace the implicit temporary stream with an explicitly ordered stream:

```python
caller_stream = torch.cuda.current_stream()
benchmark_stream = torch.cuda.Stream()
benchmark_stream.wait_stream(caller_stream)
with torch.cuda.stream(benchmark_stream):
    # Existing warmup and benchmark body.
    fn()
    ...
```

The wait must be established before the first warmup launch. This creates a GPU stream dependency without adding a host-blocking global synchronization.

A regression test should exercise both default and non-default caller streams and check data ordering, rather than only checking that the helper returns a timing value.

k6699k:
## What

- Record the caller's current CUDA stream before creating the benchmark stream.
- Make the benchmark stream wait for previously submitted caller-stream work before warmup and CUDA Graph capture.
- Add regression coverage for both default-stream and non-default-stream callers.
- Preserve the existing benchmark API and configurable `n_retries` behavior.

## Why

`do_bench_cudagraph()` previously switched to a newly created stream without establishing a dependency on the caller stream. The benchmark stream could therefore overtake pending input preparation or tensor operations, causing an asynchronous stream race and potentially producing incorrect or unstable results.

The explicit `wait_stream()` preserves the expected execution order without adding a device-wide synchronization or including the wait in the measured graph replay latency.

## Testing

- Verified the regression test fails with the original implementation.
- Verified both default-stream and non-default-stream cases pass with the fix on Hygon.
- Confirmed the required `current_stream()`, `Stream.wait_stream()`, and `CUDAGraph` APIs are available in the NVIDIA H20, MetaX, and Hygon environments.

Fixes #1127


Vincent-Xiao (dismissed):

python/triton/testing.py
```diff
@@ -103,7 +103,11 @@ def do_bench_cudagraph(fn, rep=20, grad_to_none=None, quantiles=None, return_mod
     if not isinstance(n_retries, int) or isinstance(n_retries, bool) or n_retries <= 0:
         raise ValueError("n_retries must be a positive integer")
 
-    with torch.cuda.stream(torch.cuda.Stream()):
+    caller_stream = torch.cuda.current_stream()
+    benchmark_stream = torch.cuda.Stream()
+    # Preserve work queued by the caller before switching streams for warmup.
+    benchmark_stream.wait_stream(caller_stream)
+    with torch.cuda.stream(benchmark_stream):
```
zhzhcookie, line 111:
```python
    # flagtree flagtune: Preserve work queued by the caller before switching streams for warmup.
    # with torch.cuda.stream(torch.cuda.Stream()):
    caller_stream = torch.cuda.current_stream()
    benchmark_stream = torch.cuda.Stream()
    benchmark_stream.wait_stream(caller_stream)
    with torch.cuda.stream(benchmark_stream):
```

zhzhcookie (approved):
LGTM

PR #1148, opened 2026-09-10
[CI] Refactor FlagGems test and compare actions

liuhaoyutz:
- Set PYTHONPATH to the FlagGems source directory before running tests, so the test runner can import the package directly from source without requiring it to be installed.
- Simplify baseline path handling in compare-flaggems-test action, remove the date-based baseline path construction and read the baseline directly from ${baseline-dir}/summary.json.


zhzhcookie (approved):
LGTM

PR #1150, opened 2026-09-10
[Triton] Fea/warp layout broadcast from upstream triton

Kafka-Hatsune:
## Summary

Port of [triton-lang/triton#11646](https://github.com/triton-lang/triton/pull/11646) to FlagTree, including its prerequisites and adaptations for FlagTree’s existing layout APIs and backend implementations.

Allow supported layout conversions to read broadcast copies within the current warp, avoiding unnecessary shared-memory transfers and barriers. Preserve the existing `mixedTranspositions.size() < 2` cost check and shared-memory fallback.

## Motivation

A conversion can appear to require communication across warps or CTAs when the conversion mapping selects a remote representative of a broadcast value. However, the same value may already be available within the destination warp.

For example, in the `16×1` pointer-layout regression, every source warp already contains the required elements. The previous lowering nevertheless allocated shared memory and inserted a barrier.

This change selects local broadcast copies and carries warp/CTA-dependent source-lane selection into the existing shuffle decomposition and lowering.

- example
    
    ```cpp
    tensor<16x1xptr, #src> -> tensor<16x1xptr, #dst>
    Src：
    sizePerThread = [1,1]
    threadsPerWarp = [32,1]
    warpsPerCTA = [4,1]
    order = [1,0]
    Dst：
    sizePerThread = [1,8]
    threadsPerWarp = [4,8]
    warpsPerCTA = [4,1]
    order = [1,0]
    ```
    
- The P in the SP=D convertLayout matrix is modified to
    
    ```cpp
    P                dl0 dl1 dl2 dl3 dl4 dw0 dw1
    src lane bit 0    0   0   0   1   0   0   0
    src lane bit 1    0   0   0   0   1   0   0
    src lane bit 2    0   0   0   0   0   1   0
    src lane bit 3    0   0   0   0   0   0   1
    src lane bit 4    0   0   0   0   0   0   0
    src warp bit 0    0   0   0   0   0   1   0  <- from 0 to identity matrix
    src warp bit 1    0   0   0   0   0   0   1
    ```
    
    The analysis detects zero columns corresponding to warp bits in the source layout matrix S. These source warp coordinates do not affect the logical element index, so their corresponding rows in P can be set to copy the matching destination warp bits while preserving SP = D. In this example, both source warp bits are broadcast, making the warp-to-warp submatrix of P an identity matrix.
    
    equivalently:
    
    ```
    srcLane = (dstLane >> 3) + 4 * dstWarp
    srcWarp = dstWarp
    ```
    

## What changed

- **Upstream prerequisites** — add input/output basis-mask helpers and update packed-transposition selectors to the upstream implementation preceding #11646, including selector ordering and low-register-bit permutations.
- **`LayoutUtils.h` / `LayoutUtils.cpp`** — introduce `invertAndComposeLocal`. When source warp/block basis vectors are zero, select the corresponding destination coordinates while preserving the logical element mapping.
- **Analysis `Utility.h` / `Utility.cpp`** — generalize the decomposition from `pLane` to `shuffleMap`, carrying register/lane/warp/block contributions to source-lane selection. Extend mixed-transposition metadata with separate source and destination lane bits, including one-sided exchanges.
- **Common `ConvertLayoutOpToLLVM.cpp`** — check shuffle eligibility before the shared-memory fallback and use the generalized mapping during shuffle emission. Preserve register packing, broadcast handling, and the existing diagnostic for unsupported cross-CTA transfers.
- **FlagTree API compatibility** — retain tensor-type analysis interfaces and use explicit zero padding with FlagTree’s existing `LinearLayout` API. Keep permutation checks local to analysis. The newer upstream `force_warp_shuffle` attribute and `GenericLinearEncoding` are outside this backport.
- **AMD/Metax compatibility** — adapt AMD permlane lowering to the new decomposition. Since Metax also compiles this AMD source against its own older analysis header, use the existing `USE_MACA` condition to select `pLane`/`transposition` for Metax. Metax retains its existing analysis and shuffle eligibility rules.
- **Regression coverage and provenance** — add NVIDIA pointer, AMD wave64, local-broadcast, packed-exchange, and GPU correctness coverage. Annotate modified code with upstream provenance and FlagTree adaptation explanations.

## Test plan

Verified in the H20：

- **Before/after IR comparison** for the `16×1` pointer conversion:
    
    
    | Generated code | Baseline | After |
    | --- | --- | --- |
    | Shared-memory allocation | 128 bytes | 0 bytes |
    | 32-bit shuffle instructions | 0 | 2 |
    | Barriers | 1 | 0 |
    | Shared-memory loads/stores | 1 each | 0 |
- **Build and formatting:** NVIDIA/AMD build and pre-commit passed.
- **GPU correctness:** feature validation completed with **236 passed, 244 skipped**, including all **88 newly added cases**. The 88 added cases passed again after the Metax compatibility fix.
- **Focused compiler tests:** all **5 conversion IR files** and **5 C++ test groups** passed.
- **AMD permlane:** all **24 existing gfx950 cases** compiled and produced identical LLVM IR before and after the compatibility fix.
- **Metax interface compatibility:** reproduced the missing-member compilation errors using the actual Metax analysis header; the same AMD translation unit passed syntax checking after the fix.
- **Broader IR suite:** **206 passed, 13 failed**. All 13 failures were also reproduced with the unmodified FlagTree baseline in the same environment.


zhzhcookie (approved):
LGTM

PR #1153, opened 2026-09-11
[CI][iluvatar] Add iluvatar vllm benchmark to workflow

ShawnLiuxinghong:
Add iluvatar directory and related configurations under benchmark,along with iluvatar3.6-qwen-benchmark.yaml


zhzhcookie (approved):
LGTM

Issue #1152, opened 2026-09-11
[Bug][metax][TLE] smem support is incomplete: scalar-indexed `local_ptr` asserts at LoadStoreOpToLLVM.cpp:427, smem atomics are unsupported, and smem+histogram silently corrupts data


LittleShun1214:
- Repo: https://github.com/flagos-ai/FlagTree
- Labels: `bug`, `metax`, `tle`
- Environment: flagtree main `2cd4cc6a` (wheel `0.6.0+metax.git2cd4cc6a`, 2026-09-09) on MetaX C550
- All reproducers are inlined in the Appendix; every listed kernel is self-contained (torch + flagtree only)
- Related: #742 (Add Metax TLE support), #971 (Metax TLE support local pointer)

## Summary

Metax TLE shared-memory support is only partially implemented, and the supported/unsupported split is asymmetric. Five problems, each with a minimal deterministic reproducer:

| # | Problem | Kind | Listing |
|---|---|---|---|
| 1 | Scalar-indexed `local_ptr` + offset access asserts at `LoadStoreOpToLLVM.cpp:427` | compile-time | 1, 2 |
| 2 | `tt.atomic_rmw` on smem pointers fails front-end verification | compile-time | 3 |
| 3 | `local_ptr` indices depending on a dynamic loop variable crash | compile-time | 4, 5, 6 |
| 4 | TLE smem buffer overlaps `tl.histogram`'s smem scratch (large tile) | runtime | 7, 8 |
| 5 | Masked smem load ignores `other` | wrong value (minor) | 9 |
| – | `buffered_tensor` cannot cross function boundaries | limitation | 10 |
| – | smem capacity: 32 KB + histogram scratch fits, 64 KB does not | informational | — |

All affected patterns are used by production TLE kernels on NVIDIA (FlagGems `top_k_per_row_decode` / `bin_topk`, FlagGems-vllm `persistent_topk`) and/or covered by upstream TLE tests (`python/test/tle/unit/test_tle_gpu_local_ptr.py`). The metax CI (`metax3.6-build-and-test.yml`) does not run `python/test/tle/`, so none of them are exercised on metax.

The merged metax TLE local-pointer PR #971 explicitly declares support for scalar, tensor, static and dynamic `local_ptr` indices, for masked loads and stores through local pointers, and for Triton atomic operations on shared memory. Problems 1, 2, 3 and 5 below are in direct contradiction with those declarations.

## Environment

| | |
|---|---|
[...]

PR #971, opened 2026-08-12
[Metax][TLE] Metax TLE support local pointer.

cjh9368:
## METAX backend support for TLE GPU local pointers

This patch adds METAX/MACA support for TLE local pointers. TLE kernels can now create pointers to shared-memory allocations and access the referenced data through standard Triton load/store operations.

## Supported APIs

* `tle.gpu.alloc(..., scope=tle.gpu.smem)`
  * Allocates a TLE shared-memory buffer.

* `tle.gpu.local_ptr(buffer, indices=None)`
  * Creates local pointers to a TLE shared-memory allocation.
  * Supports full-buffer views when `indices` is omitted.
  * Supports indexed 1-D and multi-dimensional views.
  * Supports scalar, tensor, static, and dynamic indices.

* `tl.load(local_ptr, mask=None, other=None)`
  * Loads data through TLE local pointers.
  * Supports masked and unmasked loads.

* `tl.store(local_ptr, value, mask=None)`
  * Stores data through TLE local pointers.
  * Supports masked and unmasked stores.

These APIs can be used in elementwise kernels, loops, and conditional control-flow regions. Local pointers can also be used together with `tle.gpu.copy` to stage global-memory data in TLE shared memory before accessing it through `tl.load` and `tl.store`.

## Lowering path

Key points for the METAX implementation:

* Local-pointer loads and stores are lowered to METAX shared-memory load/store instructions.
* The lowering detects the address space of the generated LLVM pointers and selects the shared-memory path when necessary.
* Vectorized shared-memory loads and stores are supported when pointer alignment permits.
* Shared-memory side effects are modeled for Triton atomic operations.
* The TLE local-pointer barrier insertion pass remains enabled to preserve the required memory ordering.

## Scope and limitations

This PR enables TLE local pointers for load/store-based kernels.

TLE kernels containing a `tt.dot`/`tl.dot` operation are not supported on the METAX backend yet. Such kernels are rejected during compilation. Dot-based local-pointer tests are therefore excluded from the METAX test suite for now.

PR #1164, opened 2026-09-14
[KMCompiler][Metax][TLE] Pass __MCTLE__ to TableGen and keep local_pointers buffers alive in alias analysis

cheersluvs:
## Summary

Two build/analysis fixes that make TLE shared memory (`tle.gpu.alloc` + `tle.gpu.local_ptr`) usable on the MetaX backend. Each commit fixes one problem:

| Commit | File | Problem fixed |
|---|---|---|
| `[Metax][TLE] Pass __MCTLE__ to TableGen as well` | `cmake/FlagTreeOptions.cmake` | `tt.atomic_rmw` / `tt.atomic_cas` on shared-memory pointers are rejected by the verifier |
| `[Metax][TLE] Keep shared buffers reached through mctle.local_pointers alive` | `third_party/metax/lib/Analysis/Alias.cpp` | the shared-memory allocator reuses TLE buffers that are still in use, **silently corrupting data** or faulting |

Relates to #1152: this fixes its **Problem 2** (smem atomics) and **Problem 4** (TLE buffer overlapping `tl.histogram`'s scratch). It does not address Problems 1, 3 and 5, which are in the prebuilt `metaxTritonPlugin.so` lowering.

## Problem 1: `__MCTLE__` never reaches TableGen

**Symptom.** Any atomic on a shared-memory pointer fails at TTIR verification:

```
error: 'tt.atomic_rmw' op failed to verify that ptr type matches value type
```

```python
@triton.jit
def kernel(out, N: tl.constexpr):
    i = tl.arange(0, N)
    buf = tle.gpu.alloc([N], dtype=tl.int32, layout=None,
    p = tle.gpu.local_ptr(buf)
    tl.store(p, i)
    tl.debug_barrier()
    old = tl.atomic_add(p, 7, sem="relaxed", scope="cta")   # <- verifier error
    tl.store(out + i, old)
```

**Root cause.** #971 put the shared-pointer form of the atomic constraint in `third_party/metax/include/triton/Dialect/Triton/IR/TritonOps.td` behind `#ifdef __MCTLE__` (`getPointeeType`), with the original constraint in the `#else` branch (`getPointerTypeSameShape`, which hard-codes address space 1; shared-memory pointers are address space 3). The shared-memory effects in the same file are guarded the same way. `cmake/FlagTreeOptions.cmake` defines `__MCTLE__` with `add_definitions`, which applies to the C++ compiler only and never reaches `mlir-tblgen`, so the generated verifier always used the `#else` branch.

**Fix.** Append `-D__MCTLE__` to `LLVM_TABLEGEN_FLAGS` next to the existing `add_definitions`, mirroring how `__TLE__` is already removed from `LLVM_TABLEGEN_FLAGS` a few lines below.

## Problem 2: shared-memory buffers reached only through `local_ptr` look dead

**Symptom.** No compile error; results are wrong or the kernel faults, depending on layout. Typical observations:

- two buffers written through `local_ptr` and read back after both are written: the first buffer reads back the second one's data, and `metadata.shared` covers only one buffer (e.g. 2048 B for two 512 x int32 buffers);
- a TLE buffer followed by a cross-warp reduction or `tl.histogram`: the reduction scratch lands inside the buffer (#1152 Listing 7);
- a real 4-round radix top-k kernel: the LLVM IR placed reduction scratch at `@global_smem+1..7`, inside a 2048 x int32 histogram buffer, and a 512 x float32 buffer got no offset of its own (`metadata.shared` = 18484 B, 2012 B less than the declared buffers). The kernel returned duplicate indices on some inputs and hit `memory violation(0x4)` on others.

**Root cause.** The metax backend builds its own `third_party/metax/lib/Analysis/Alias.cpp`, which does not have the `local_pointers` handling that `lib/Analysis/Alias.cpp` has under `#ifdef __TLE__`. Pointers produced by `mctle.local_pointers`, and `tt.splat` / `tt.broadcast` / `tt.addptr` chains over them, therefore carry no alias. `Allocation` then ends the buffer's lifetime at `local_pointers` and reuses its bytes, even though later `tt.load` / `tt.store` / `tt.atomic_rmw` still access it through those pointers.

**Fix** (under `#ifdef __MCTLE__`):

1. `mctle.local_pointers` results alias their source memdesc (operand 0).
2. For any other op, each pointer-like result joins the aliases of all operands, **result by result**. A pointer returned as a later result, e.g. the second result of `inline_asm_elementwise`, keeps its buffer alive too. Non-pointer results keep the entry state, and the existing "unknown operation creating memory descriptor" assertion is kept for them.

The join is a conservative over-approximation: it can only extend a buffer's lifetime, never shorten it. Without `__MCTLE__` the file is unchanged.

## Testing

Environment: MetaX C550, MACA 3.7.1, torch 2.8.0+metax3.7.1.4, triton 3.6.0 (FlagTree, `FLAGTREE_BACKEND=metax`, `BUILD_MCTLE=ON`), prebuilt `metaxTritonPlugin.so` v0.6.2.

| Check | Without the fixes | With the fixes |
|---|---|---|
| `tl.atomic_add` on a `local_ptr` pointer (full view, scalar + offsets, masked, scatter) | verifier error | correct |
| Two buffers through tensor-index `local_ptr` (N=512, 4 warps) | wrong values, `shared` 2048 B | correct, `shared` 4096 B |
| Two buffers through scalar `local_ptr` + `tt.addptr` (N=256, 4 warps) | first buffer wrong, `shared` 1024 B | correct, `shared` 2048 B |
| Pointer returned as the 2nd result of `inline_asm_elementwise` | wrong (also with a result-0-only alias rule) | correct, `shared` 2048 B |
| smem round-trip + `tl.histogram` (#1152 Listing 7 shape) | corrupted | correct |


cjh9368 (approved):
Please add automated regression tests for these fixes. The PR description includes manual test results, but no tests are added in the diff, and buffer lifetime bugs like this are easy to reintroduce.
At a minimum, please cover:
- Verification of atomic_rmw and atomic_cas on shared-memory pointers.
- Two simultaneously live shared-memory buffers receiving non-overlapping allocations.
- A shared buffer remaining live across a histogram or reduction that uses shared-memory scratch.
- A multi-result operation whose pointer result is not the first result.
Verifier, alias-analysis, and allocation checks can be implemented as compiler-level tests without requiring GPU hardware.

cheersluvs:
Thanks for the review. Added `third_party/metax/python/test/unit/language/test_mctle_shared_memory.py` (7ceab88b2), which covers the four cases you listed:

- `tt.atomic_rmw` and `tt.atomic_cas` on shared-memory pointers: the op is checked in TTIR on an address-space-3 pointer (i.e. verification passed), and the values are checked on the device;
- two simultaneously live buffers, through tensor-index `local_ptr` and through a scalar `local_ptr` + `addptr` chain: `metadata.shared` must cover both buffers, and neither buffer may be overwritten by the other;
- a buffer kept live across `tl.histogram` and across cross-warp `tl.sum` / `tl.cumsum`;
- a pointer returned as the second result of `inline_asm_elementwise`.

Results on a C550:

| build | result |
|---|---|
| this PR | 7 passed |
| this PR with the earlier result-0-only alias rule | `test_mctle_pointer_as_second_result_keeps_buffer_live` fails (`shared` 1024 B, need 2048 B), other 6 pass |
| published wheel without mctle | 7 skipped |

The module is skipped when the metax backend is built without mctle.

Vincent-Xiao (approved):

Vincent-Xiao:
LGTM

zhzhcookie (approved):
LGTM

PR #1166, opened 2026-09-14
[QC][Nvidia] Barrier & Sync Optimization

monellz:
# Optimization 
## 1. Membar: stop emitting redundant `bar.sync` around mbarrier ops

`lib/Analysis/Membar.cpp`, `third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/TritonGPUToLLVM.cpp`

- **Avoid redundant bar.sync around mbarrier arrive/wait** — the all-shared-memory
  dependency of a non-participant arrive is only needed to order the arrive against
  prior accesses. It is now checked with a transient probe instead of being joined into
  the analysis state (where it forced a rendezvous at every later access), and
  `canSkipBarSync` exempts arrive→wait pairs, which the hardware already orders.
  Removes one `bar.sync` per slot in mbarrier ping-pong loops.
- **Skip bar.sync between consecutive mbarrier arrives** — the barrier before the first
  arrive already collected the group, so back-to-back arrives need no rendezvous in
  between. Removes one `bar.sync` per iteration in loops that release several slots
  together (warp-specialized GEMM consumers, for example).
- **Fix interval refinement for block arguments** — correctness fix in the same area: a
  block argument was treated as a view at the buffer base, collapsing captured sub-views
  onto one interval, so real hazards reached through mixed paths could be missed while
  disjoint slots conflicted spuriously. Warp-specialize partition arguments are now
  resolved through the parent's captures; other block arguments fall back to the whole
  buffer.

## 2. TLE: schedule TMA store waits across control flow

`third_party/tle/dialect/lib/Transforms/TleScheduleTmaStoreSync.cpp`

`TleScheduleTmaStoreSync` drained every block: each TMA store group was committed and
immediately waited for, so no two stores were ever in flight. It now tracks pending
groups through branches and loop back edges and waits at source-reuse or lifetime
boundaries. A store loop keeps up to 8 groups in flight (`cp.async.bulk.wait_group.read 7`
inside the loop, full drain at the exit) instead of `wait_group.read 0` every iteration.

## Measured impact

H20 (78 SM), driver 590.48.01, PyTorch 2.11.0+cu130, Triton/FlagTree 3.6.0,
FlagGems `34bd6d6`. The operators and the TMA store repro are timed with
`do_bench_cudagraph(rep=..., n_retries=10, median)`, the two membar repros with CUDA
events over five timed launches after warm-up. Both arms run in the same process against
the same buffers, alternating order, with an unaffected kernel timed alongside as a drift
control. "Latency reduction" is `1 - new/old`, not a throughput gain.

### FlagGems `fp8_einsum` (w8a8 block-fp8 bmm) — part 1

Hopper warp-specialized TLE kernel. Against the baseline, the only textual difference in
the generated PTX of the selected kernel is one removed `bar.sync` (static count 7 → 6,
SASS `BAR.SYNC` 20 → 19; inside the K loop 3 → 2); QGMMA and
`WARPGROUP.DEPBAR` counts are unchanged, so this is pure synchronization overhead.

| shape (b, h, r), D=1024 | baseline | this branch | latency reduction |
|---|---:|---:|---:|
| (8192, 8, 4096)  | 2.1765 ms | 2.1402 ms | **1.67 %** |
| (32768, 8, 4096) | 8.6064 ms | 8.4986 ms | **1.25 %** |
| (8192, 16, 7168) | 7.4638 ms | 7.3426 ms | **1.62 %** |

### FlagGems `aten.fill_.Scalar`, 67,108,864 elements, BF16 — part 2

Shape taken from `benchmark/models_shapes/qwen25.yaml` (`fill_`), scalar from
`benchmark/test_fill.py`. The library's Hopper `fill_` uses plain `tl.store`, so the
comparison uses a TLE fill kernel that writes one contiguous 1 KiB tile per TMA store
(`rows=4, width=128, waves=4, warps=4`, grid 312); the same Python source and the same
configuration are compiled by both compilers, so the difference is the store schedule
alone.

| variant (value = 1.25) | latency |
|---|---:|
| TLE kernel, old schedule | 45.183 µs |
| TLE kernel, this branch | **37.209 µs** (-17.6 %) |
| FlagGems Hopper `fill_` (library) | 40.303 µs (this branch is 7.7 % faster) |


zeroherolin:
Thanks to @monellz 

I have reviewed the code and run it on H20\. The Membar part is fine\. The TMA store scheduling rewrite has a correctness regression; it needs a fix and tests before merging\.

## 1\. Membar: no issues

Turning the elected arrive into a transient probe and letting arrive→wait / arrive→arrive skip `bar.sync` are consistent with each other; the block\-argument interval change is a correctness fix\. Built the base commit and the PR separately and compiled the warp\-specialized GEMM in `test_tle_ws_tma_gemm`: static `bar.sync` in the PTX drops from 19 to 14, named barriers stay at 9, tests pass\. **One side effect: memdescs carried through `scf.for` iter\_args now fall back to the whole allocated interval, so pipelined kernels may get extra barriers\. Please check\.**

## 2\. TMA store scheduling: correctness regression, fix before merging

### 2\.1 The source buffer can be reused before the wait

The new schedule lets pending groups flow out of `scf.if` / `scf.for` \(`isControlFlow` returns early, so region terminators no longer drain\), while reuse detection only looks at alias roots\. If the source buffer is allocated inside the region, it is dead once the region exits; Allocation hands its offset to the next `local_alloc`, and the TMA engine reads the newly written data\. A `bar.sync` cannot prevent this\.

Minimal repro \(pass 3 for `flag` so the `if` is not constant\-folded\):

```python
@triton.jit
def k(out_desc, sink, flag, XB: tl.constexpr, YB: tl.constexpr):
    pid = tl.program_id(0)
    if flag != 0:
        src = tle.gpu.alloc([XB, YB], tl.float32, init_value=tl.full((XB, YB), 1.0, tl.float32))
        for g in tl.static_range(8):
            tle.gpu.copy(src, out_desc, [XB, YB], [pid * XB, g * YB])
    # src is dead here; dst may get the same shared offset
    dst = tle.gpu.alloc([XB, YB], tl.float32, init_value=tl.zeros((XB, YB), tl.float32))
    rows = tl.broadcast_to(tl.arange(0, XB)[:, None], (XB, YB))
    cols = tl.broadcast_to(tl.arange(0, YB)[None, :], (XB, YB))
    tl.store(sink + rows * YB + cols, tl.load(tle.gpu.local_ptr(dst, (rows, cols))))
```

XB=128, YB=32, grid 64, num\_warps 4, 50 runs per variant\. "Before" is the base commit 5eda9956b; "wrong" means zeros show up in the output\.

|Variant|Wrong runs before / 50|Wrong runs with this PR / 50|
|---|---|---|
|Source buffer allocated inside the `if` \(above\)|0|50|
|Same code without the `if`|50|50|
|Without the `if`, and `src` is read again after `dst`|0|0|

The old implementation does not model offset reuse either; the full drain at region exits merely hid the cross\-region case\. Keeping `src` alive past `dst` makes the result correct, so the mechanism is offset reuse\.

Fix: make the wait a use of the source buffer \(attach the source memdesc as an operand of the wait, or insert `local_dealloc` after it\) so that liveness extends to the wait; at minimum, wait by age at region terminators for pending roots defined inside that region\. Please also add a GPU regression test for this scenario and lit cases with `scf.if` / `scf.for`; all 4 existing lit cases are straight\-line blocks\.

### 2\.2 The benefit is narrow and depends on warp configuration

A write to the same alias root triggers wait\(age\), and all slot views produced by `memdesc_index` share one root, so a loop that rewrites the source buffer every iteration, the usual GEMM epilogue, still waits with 0 pending every iteration\. "8 groups in flight" only happens when the source buffer is loop\-invariant, as in fill\. For load→compute→store loops there is a small but real gain: the wait moves from the end of the iteration to just before the next iteration rewrites the source, confirmed on the elementwise kernel\.

The gain also depends on warp configuration: on the same H20 with about 67M BF16 elements and an 8x64 tile, 1 warp is 9\.3% faster while 4 warps are 9\.3% slower\. **Please add a tile / warps / pending\-groups performance matrix; a fixed 8 groups is not a general win\.**

monellz:
Add four commits. 

**`c32721780`**: a group now completes before the region its source was allocated in exits, and all pending groups drain before any new shared allocation.

**`5d1c3c042`**: the number of groups in flight is a pass option, `max-pending-groups`, defaulting to 1.

**`a9e88567d`**: a loop-carried block argument resolves to the union of its init and yielded views instead of the whole allocation.

**`1bb9d5d6f`**: every line this PR adds to `Membar.cpp` and the NVIDIA lowering is behind `__TLE__`, so a build without TLE keeps the upstream barriers.

## Test files

- `python/test/tle/integration/test_tle_tma_store_dataflow.py` (new) — store lifetime on
  GPU, including `test_store_source_offset_reuse[straight_line|in_region]`, which fails
  before the fix and on `main`. Needs a fresh Triton cache per arm, or the second arm
  reuses the first arm's cubin.
- `third_party/tle/test/Analysis/test-membar-static-views.mlir` (new) — disjoint slots,
  same slot, and both carried through `iter_args`.
- `third_party/tle/test/GPU/test_tle_schedule_tma_store_sync.mlir` (extended) — four
  control-flow cases (back edge, source allocated in an `scf.if`, source outside it, new
  allocation), each run once and twice, plus a second RUN line at `max-pending-groups=8`.
- `third_party/tle/dialect/include/Transforms/Passes.td` — the option, clamped to 1…8.

## Tile / warps / pending-groups matrix

H20, 67,108,864 BF16 elements, `waves=4`, `do_bench_cudagraph` median of three
alternating rounds; every limit compiled in one process and checked to emit different
PTX. `cap` = groups in flight; `cap=1` keeps the new wait placement but overlaps
nothing with other groups. Times in µs.

| tile | bytes | warps | cap 1 | cap 2 | cap 4 | cap 8 | cap 8 vs cap 1 |
|---|---:|---:|---:|---:|---:|---:|---:|
| 8x64  | 1024 | 1 | 41.45 | 41.86 | 41.69 | 41.69 | +0.59 % |
| 8x64  | 1024 | 4 | 37.22 | 37.41 | 37.40 | 37.41 | +0.51 % |
| 8x64  | 1024 | 8 | 56.90 | 54.82 | 54.82 | 54.84 | -3.62 % |
| 4x128 | 1024 | 1 | 41.82 | 41.66 | 41.71 | 41.83 | +0.02 % |
| 4x128 | 1024 | 4 | 37.21 | 37.17 | 37.20 | 37.20 | -0.01 % |
| 4x128 | 1024 | 8 | 57.01 | 54.97 | 55.00 | 54.91 | -3.69 % |
| 2x256 | 1024 | 1 | 41.68 | 41.70 | 41.85 | 41.50 | -0.42 % |
| 2x256 | 1024 | 4 | 37.23 | 37.28 | 37.28 | 37.29 | +0.15 % |
| 2x256 | 1024 | 8 | 56.93 | 54.82 | 54.84 | 54.82 | -3.70 % |
| 8x128 | 2048 | 1 | 42.27 | 42.17 | 42.06 | 42.15 | -0.29 % |
| 8x128 | 2048 | 4 | 40.99 | 43.59 | 43.53 | 43.48 | **+6.07 %** |
| 8x128 | 2048 | 8 | 43.39 | 40.06 | 40.06 | 40.06 | **-7.68 %** |

monellz:
Added in `8f76fd9c0`, rebased on `c9372e9a6`. A kernel can now state how many TMA store groups it may
keep in flight:

```python
@triton.jit
def kernel(desc, ...):
    tle.gpu.tma_store_pending(8)          # 1..8, compile time, optional
    staging = tle.gpu.alloc([BM, BN], tl.float32, scope=tle.gpu.smem)
    for i in tl.range(0, n):
        tle.gpu.copy(staging, desc, [BM, BN], [i * BM, 0])
```

monellz:
Now it uses #hint instead of new api.

```python
for i in tl.range(...):
    tle.gpu.copy(...)  #@hint: tma_store_pending=8
```

python/triton/experimental/tle/language/gpu/core.py
```diff
@@ -1064,7 +1124,13 @@ def copy(
 
         Masked global -> local copy with zero fill:
             tle.copy(global_ptrs, local_buf, [64, 128], mask=valid)
+
+        TMA store that may keep several groups in flight:
+            for i in tl.range(0, n):
+                tle.copy(local_buf, tma_desc, [64, 128], [i * 64, 0])  # @hint: tma_store_pending=8
     """
+    _mark_tma_store_pending(flagtree_hints, _semantic, _generator)
```
sunnycase: Could we restrict this hint to actual local-to-global TMA copies?

`_mark_tma_store_pending()` is called unconditionally for every `copy()` invocation, while `NvidiaHintHandler` forwards hints for all copy calls. Therefore, placing the documented `tma_store_pending=8` hint on a global-to-local TMA copy or a normal pointer copy also raises the module-wide bound used for unrelated TMA stores. This may not cause incorrect code, but it can change wait placement, shared-memory liveness, and resource usage unexpectedly.

Please consider moving this call after the direction/type checks and accepting it only when the operation is a local-to-global TMA copy (or rejecting the hint for other directions). A regression test for a hint on a global-to-local copy would also be helpful.

lib/Analysis/Membar.cpp
```diff
@@ -206,12 +226,45 @@ static std::optional<StaticAccessView> getStaticMemDescView(Value value) {
...
+    // the buffer base.
+    if (auto partitions = dyn_cast<ttg::WarpSpecializePartitionsOp>(
+            arg.getOwner()->getParentOp()))
+      return getStaticMemDescView(
+          partitions.getParentOp().getExplicitCaptures()[arg.getArgNumber()],
+          visited);
+    auto loop =
+        dyn_cast_or_null<LoopLikeOpInterface>(arg.getOwner()->getParentOp());
+    if (!loop || !visited.insert(value).second)
+      return std::nullopt;
```
sunnycase: Could we preserve the loop-init view when the yielded view is derived from the loop-carried argument?

For example, with a pattern such as `%sub = ttg.memdesc_subslice %cur [...]` followed by `scf.yield %sub`, the recursive lookup of `%cur` while analyzing `%sub` reaches this `visited` check and returns `nullopt`. `mergeStaticViews()` then fails, so the analysis falls back to the whole allocation. This can introduce extra barriers even when `%cur` and another accessed slot are statically disjoint.

Please consider resolving the yielded view using the loop-init view as the block-argument context, while retaining cycle protection for genuine cycles. A regression case such as `carried_disjoint_derived_view` in `third_party/tle/test/Analysis/test-membar-static-views.mlir` would help preserve this precision.

python/triton/experimental/tle/language/gpu/core.py
monellz: Fixed. Now it will check the direction and apply only to local-to-global tma copies.

lib/Analysis/Membar.cpp
monellz: Fixed. The newly added test `carried_disjoint_derived_view` generates 2 barriers before and 1 after the fix.

PR #1167, opened 2026-09-14
[CI][TLE-Distributed] Remove USE_FLAGCX variable from workflow

Galaxy1458:
Removed USE_FLAGCX environment variable from the workflow.


zhzhcookie (approved):
LGTM

PR #1169, opened 2026-09-14
[HCU][TLE] Allow extract_tile in the TLE whitelist

lzllx123:
## Summary
Add `extract_tile` to the TLE whitelist in the HCU backend (`third_partycu/backend/tle_supported.py`), so that kernels using `extract_tile` are allowed to be lowered/executed through the TLE path instead of being rejected.


sunnycase (approved):
LGTM

PR #1170, opened 2026-09-14
[TLE] Expose the sync scope/memory order in tle.signal/tle.signal_wait frontend

legilla:
also tweaked the barrier ops to use the added attrs


Galaxy1458 (approved):
LGTM

PR #1174, opened 2026-09-15
[TLE][Ascend][XCS] Add writable attr for tile_to_tensor

finneyyan:
Lower tile.to_tensor with the writable attr to bufferization.to_tensor restrict writable so custom ops write results back in place, instead of One-Shot Bufferization materializing an unplannable UB allocation that becomes a device-side malloc and breaks the dataflow.


zhzhcookie (approved):
LGTM

sunnycase (approved):
LGTM

PR #1187, opened 2026-09-16
[CI] Add pre-command input to run-flaggems-test action

liuhaoyutz:
Introduce an optional `pre-command` input to run-flaggems-test action, allow each backend to perform its own required setup before running FlagGems tests.


zhzhcookie (approved):
LGTM

PR #1190, opened 2026-09-16
[Benchmark][PPU] Update ppu benchmark results

ShawnLiuxinghong:
Test results of different Flagtree versions on the 890p and 810


zhzhcookie (approved):
LGTM

PR #1192, opened 2026-09-16
[CI] Add backend-specific setup interface to gems templates

liuhaoyutz:
Allow each backend to perform backend-specific setup before running FlagGems tests by introducing two optional inputs to the gems-baseline and gems-test workflow templates.

- flaggems-pre-command: multiline script forwarded to the run-flaggems-test action and executed right before the test runner (e.g. source env, set LD_LIBRARY_PATH).
- env-setup-command: multiline script run in the Setup Environment step; only its `echo KEY=VALUE` output is appended to GITHUB_ENV so the variables are visible to all later steps (build, test, compare).


zhzhcookie (approved):
LGTM

Issue #1194, opened 2026-09-16
[BUG][Ascend] Launcher workspace size overflows in 32-bit multiplication before uint64_t assignment


YaooXu:
## Summary

The Ascend generated C++ launcher computes the workspace byte count using 32-bit unsigned multiplication, then assigns the already-overflowed result to `uint64_t`. A launch requiring 4.5 GiB can therefore allocate only 0.5 GiB, leading to out-of-bounds device workspace accesses.

This is still present at the currently checked Ascend branch head, `triton_v3.5.x` @ `1aaab9566f011f3050da0f8c69cd7a1d773a9436`:

[third_party/ascend/backend/driver.py, lines 858–862](https://github.com/flagos-ai/FlagTree/blob/1aaab9566f011f3050da0f8c69cd7a1d773a9436/third_party/ascend/backend/driver.py#L858-L862)

```cpp
uint32_t blockNum4Workspace = gridX * gridY * gridZ;
// ... pre_launch hook ...
uint64_t totalWorkSpaceSize = {workspace_size} * blockNum4Workspace;
```

For example, `{workspace_size}` becomes the decimal literal `147456`. Its multiplication with `uint32_t blockNum4Workspace` takes place in 32 bits. The `uint64_t` destination does not widen the multiplication operands.

## Minimal reproducer — no NPU required

Save as `workspace_overflow.cpp`:

```cpp
#include <cassert>
#include <cstdint>
#include <iostream>

int main() {
    uint32_t blocks = 32768;
    uint64_t bad4 = 131072 * blocks;
    uint64_t good4 = static_cast<uint64_t>(131072) * blocks;
    uint64_t bad45 = 147456 * blocks;
    uint64_t good45 = static_cast<uint64_t>(147456) * blocks;
    uint64_t bad6 = 196608 * blocks;
    uint64_t good6 = static_cast<uint64_t>(196608) * blocks;
    std::cout << "4 GiB: " << bad4 << " vs " << good4 << '\n';
    std::cout << "4.5 GiB: " << bad45 << " vs " << good45 << '\n';
    std::cout << "6 GiB: " << bad6 << " vs " << good6 << '\n';
    assert(bad4 == 0 && good4 == 4294967296ULL);
    assert(bad45 == 536870912 && good45 == 4831838208ULL);
    assert(bad6 == 2147483648ULL && good6 == 6442450944ULL);
    blocks = 16384;
    uint64_t small_bad = 147456 * blocks;
[...]

PR #1195, opened 2026-09-16
[BACKEND] Fix Ascend workspace size overflow

ZQ-Struggle:
## Summary

Related to #1194.

Widen the per-block workspace size to `uint64_t` before multiplying it by the logical block count in the Ascend launcher.

The generated integer literal and `uint32_t blockNum4Workspace` currently produce a 32-bit unsigned multiplication. Assigning the result to `uint64_t` happens after overflow: for example, `147456 * 32768` evaluates to 512 MiB instead of 4.5 GiB, causing workspace under-allocation.

## Changes

- Cast an operand before multiplication so the workspace byte calculation uses 64-bit arithmetic.
- Add host-side regressions that compile and execute the arithmetic extracted from the actual generated launcher, without multi-GiB device allocations.
- Cover below/at/above 4 GiB, the reported 4.5 GiB and 6 GiB cases, 8 GiB, multidimensional grids, and the no-workspace path.
- Exercise both torch_npu and MindSpore allocation-code generation, task-queue and auto-blockify settings, and `-O0`/`-O2` compilation.

## Validation

Validated in an Ascend.

- The new regression fails before the fix and passes after it: **20 passed**.
- Repository pre-commit checks, including Ruff, YAPF, and mypy, pass for the changed files.
- An independent plain-Triton attention/checksum probe passed in two fresh processes on Ascend 910B4-1 with 32768 logical blocks and 5368840192 bytes of workspace (about 5 GiB). The allocator peak increase was 5368840704 bytes, all 32768 checksums were finite, and their maximum absolute error against the CPU reference was zero.

The NPU probe used the patched generated launcher with the existing native build and `enable_auto_blockify=False`. It was not a rerun of the reporter's original attention workload or a full fresh native build of upstream HEAD.


zhzhcookie (approved):
LGTM

PR #1206, opened 2026-09-17
[CI] Pick free torchrun master_port for nvshmem tutorials

i3wanna2:
Avoid EADDRINUSE on shared self-hosted runners when default 29500 is taken, matching the existing TLE unit shell-script port probe (+2) pattern.


zhzhcookie (approved):
LGTM

PR #1209, opened 2026-09-17
[BUILD] Port #1082 to 0.7.0-rc2-triton3.6 (plumb the wheel version through the build scripts)

shiptux:
Squashed port of #1082 onto the `0.7.0-rc2-triton3.6` integration branch.

Includes one extra commit aligning the packaging version with the **0.7.0** line this branch ships (#1082 targets the 0.6.1 line). Without the version plumbing fix on this branch, wheels built from the 2.2 tag fall back to a hardcoded version.

Per the release process the 2.2 final tag is cut on the release branch, and the packaging workflows run from the files present on the tagged commit — so this change has to be on this branch to take effect for 2.2. Same content as #1082; suggest reviewing together.


shiptux:
Rebuilt on the current rc2 branch after #1204, #1207 and #1208 landed there — the earlier conflict was against those. Same resolution as #1082 on main: the build-environment knobs and the changelog-derived `WHEEL_VERSION` are both kept. The version-alignment commit (0.6.1 → 0.7.0) follows on top, so packages built from the 2.2 final tag carry the branch's release-line version.

PR #1211, opened 2026-09-17
[BACKEND][xpu] Configurable GPUTarget backend string via TRITON_XPU_TARGET_BACKEND (default "cuda")

dongjibin1996:
## Summary

Kunlunxin XPU stack reported `GPUTarget("xpu", device_model, 1)`, colliding with the Intel XPU backend convention: downstream projects key device dispatch on `target.backend` and map `"xpu"` to **Intel** (e.g. sglang `fla/utils.py` `_check_platform()`), so Kunlun hardware was misrouted into Intel code paths.

This PR makes the XPU target backend string **configurable** and defaults it to `"cuda"`, matching the stack's torch presentation (torch.xmlir exposes the XPU under the `torch.cuda` interface):

- `XPUDriver.get_current_target()` reads `TRITON_XPU_TARGET_BACKEND` (default `"cuda"`). `"xpu"` restores the pre-migration Intel-xtriton string; any custom value (e.g. `"kunlun"`) also works.
- `XPUBackend.supports_target` accepts the configured string plus `"cuda"`/`"xpu"` aliases, so `CompiledKernel` metadata cached under any prior setting still deserializes and routes here. Only one backend is ever discovered in this stack (FLAGTREE_BACKEND restriction), so the wide alias set cannot cross-route.
- Vendored `compile()` stack-overflow retune guard matches both strings.
- Vendored `semantic._has_native_tma()` pinned to `False` (XPU has no native TMA; previously False only because the backend string was "xpu" — this prevents accidental nvidia descriptor lowering).
- `_internal_testing.is_xpu()` tracks the same accepted set.

Unchanged by design: backend package name `triton.backends.xpu`, `FLAGTREE_BACKEND` marker, TLE raw dialect `"xpu"`, `options.backend_name`, inductor glue (device_type "xpu" still routes via the supports_target alias), `tl.extra.*` static modules.

## Verification

Built and smoke-tested on P800 OAM (KL3), driver 5.18.0.0, torch.xmlir stack:

- `get_current_target()` → `GPUTarget(backend='cuda', arch=3, warp_size=1)`; `make_backend` routes to `XPUBackend`
- `TRITON_XPU_TARGET_BACKEND` = unset / `xpu` / `kunlun` all verified: target string honored, routing OK, legacy `cuda`/`xpu` cache routing OK under all three
- End-to-end triton kernel (compile + launch + numeric allclose) passes in both default (cuda) and override (xpu) modes
- `is_cuda()` flips to True as intended; `cuda_capability_geq(8)=False` (arch=3, deterministic); `_has_native_tma()=False`

## Notes for reviewers

- Switching the string switches triton cache namespaces (one-time recompile per setting).
- Ecosystem code seeing `"cuda"` will take NVIDIA-default branches; `arch` remains `device_model` (3 for KL3), so capability-gated NVIDIA features (>= sm80/90) stay off. Teams wanting the old behavior set `TRITON_XPU_TARGET_BACKEND=xpu`.


zhzhcookie (approved):
LGTM

PR #1214, opened 2026-09-18
[CI] Add Ascend3.5-910C FlagGems baseline and test workflows

liuhaoyutz:
Add ascend3.5-910c-gems-baseline.yml and ascend3.5-910c-gems-test.yml to generate FlagGems baseline and do FlagGems test for the ascend3.5-910c backend.


zhzhcookie (approved):
LGTM

PR #1216, opened 2026-09-18
[TLE] Allow large node transfers while limiting surviving tensors

lzllx123:
## Summary

- Raise the shared Python and Triton IR tensor construction limit from `2^20` to `2^25` to support larger node transfer expressions.
- Extend `TleFuseNodeRemoteTransfers` to enforce the original `2^20` limit on surviving static tensors after fusion and dead-address cleanup, including ordinary kernels and reused offsets/masks.


Galaxy1458 (approved):
LGTM

PR #1238, opened 2026-09-20
[TLE][Ascend][XCS]: Rename from tle.dsa.ascend.raw to tle.raw

finneyyan:


i3wanna2:
Thanks for the contribution. I reviewed the custom-op implementation and have two concerns about the current design:

- The same `custom` interface currently represents two different use cases: reusable backend extensions and kernel-specific implementations. This blurs whether an operation is a stable API and may encourage users to add arbitrary global extensions.
- Operations are identified only by strings in a process-global registry. This may introduce naming conflicts and make the API less discoverable and harder to validate. It also makes aliasing difficult, even though real-world use cases may require multiple public names or backend-specific implementations for the same operation.

I suggest separating the two use cases:

1. Reusable backend operations should remain under `triton.language.extra.cann.extension` and be exposed as named functions, for example:

    ```python
    triton.language.extra.cann.extension.index_select(...)
    ```

    The named function can still use the existing custom-op lowering internally.

2. Kernel-specific source or bitcode should use the existing TLE Raw abstraction through `@dialect(...)` from `triton.experimental.tle.raw`. A CANN runtime could be added under `python/triton/experimental/tle/raw/cann/`. Source implementations can lower to `tle.dsl_region`, while bitcode implementations could lower through a reused or extended `tle.extern_call`.

    The `dialect` function can be extended with CANN-specific parameters such as `format` and structured `pipeline` metadata. 

    ```python
    @dialect(
        name="cann",
        format="bitcode",
        file=Path(__file__).parent / "sort_topk.bc",
        extern_func_name="custom_sort_1d_topk_proposals_float",
        pipeline={"core": "vector", "pipe": "PIPE_V"},
    )
    def sort_topk(*args, **kwargs):
        ...

    @triton.jit
    def kernel(src, tmp, out, K: tl.constexpr):
        # args[2] (`out`) is the output/aliased operand.
        out = tle_raw.call(
            sort_topk,
            [src, tmp, out, K],
            output_indices=[2],
        )
    ```

If this direction works for you, I can first backport the TLE Raw implementation from `main` to `triton_v3.5.x`, so the CANN Raw runtime can be developed against the same interface.

PR #1240, opened 2026-09-20
fix(amd): CanonicalizePointers scf.if fat-ptr merge + AtomicCAS non-integer bitcast

zhangnju:
## Summary

Two AMD backend fixes surfaced while running FlagGems operator accuracy tests on W7900 (gfx1100). Both are ports of the corresponding handling in upstream Triton.

- **CanonicalizePointers**: merge the then/else fat-pointer attributes of an `scf.if` (via `FatPtrAttrs::intersect` — `canNarrow = lhs && rhs`, keep `smallTensorBase`/attributes only when both arms agree) instead of asserting they are equal. FlagGems `cat`/`concatenate`/`div_tensor` kernels legitimately produce `scf.if` arms whose fat pointers differ in `canNarrow`, which tripped the strict equality assert and crashed the pass (`PassManager::run failed`) on assert-enabled LLVM builds.

- **AtomicCAS**: `llvm.cmpxchg` requires signless-integer/pointer operands, so bitcast non-integer element types to an integer of the same width before the cmpxchg and bitcast the result back; also extract field `0` of the returned `{value, success}` pair. This fixes fp16 `tl.atomic_cas` (e.g. `scatter_reduce` with `reduce="multiply"`).

## Test plan

- [ ] `cat` / `concatenate` / `div_tensor` accuracy on W7900 (previously `PassManager::run failed`).
- [ ] `scatter_reduce` / `scatter_reduce_` prod on W7900 (previously the cmpxchg operand-type error).


zhangnju:
### Reviewer notes — context & upstream relationship

Both fixes were surfaced while running FlagGems operator accuracy tests on an AMD W7900 (gfx1100), and both touch only the AMD backend. Reference source: upstream Triton **3.6.0** (same line FlagTree is built on). Note that 3.6.0 is exactly the version that carries these defects — the changes are *not* a line-for-line copy from 3.6.0; they follow the canonical patterns used elsewhere in upstream and the underlying LLVM constraints.

---

**1) `CanonicalizePointers` — merge scf.if fat-ptr attrs instead of asserting**

- File: `third_party/amd/lib/TritonAMDGPUTransforms/CanonicalizePointers.cpp`, `ConvertSCFIfOp` + `FatPtrAttrs`.
- Symptom: `PassManager::run failed` crash on **assert-enabled LLVM builds**. Triggered by FlagGems `cat` / `concatenate` / `div_tensor` kernels, whose scf.if arms legitimately yield fat pointers with different `canNarrow`.
- Root cause: upstream 3.6.0 only guards the then/else equality with a `#ifndef NDEBUG` assert and then propagates the **then**-arm attrs. Release builds never check it, but our assert-enabled build hits the assert and aborts.
- Fix: add `FatPtrAttrs::intersect` and merge the two arms conservatively — `canNarrow = then && else`, keep `smallTensorBase`/attributes only when both arms agree — instead of asserting equality. This is a runtime-safe generalization of the debug-only check, not just tolerating it in release.

**2) `AtomicCAS` — bitcast non-integer element types around cmpxchg**

- File: `third_party/amd/lib/TritonAMDGPUToLLVM/LoadStoreOpToLLVM.cpp`, `AtomicCASOpConversion`.
- Symptom: `'llvm.cmpxchg' op operand must be signless integer or LLVM pointer type, but got 'f16'`. Triggered by fp16/bf16 `tl.atomic_cas`, e.g. FlagGems `scatter_reduce(reduce="multiply")` (implemented via a CAS loop).
- Root cause: `llvm.cmpxchg` only accepts signless-integer/pointer operands, but 3.6.0's AMD AtomicCAS feeds `f16`/`bf16` straight in. Additionally the **tensor** path extracts the loaded value with `extract_val(..., i)` (loop counter) instead of field 0.
- Fix: for non-integer element types, bitcast cmp/val to a same-width integer before the atomic and bitcast the loaded result back; extract field **0** of the `{value, success}` pair (aligning the tensor path with the scalar path). The bitcast idiom mirrors the sibling `AtomicRMWOpConversion` handling in the same file.

---

**Scope / risk:** both changes are AMD-only (2 files) and preserve existing behavior for integer types (`valueElemIntTy` is null → no-op) and for scf.if with matching arms. Verified on gfx1100: `cat` / `concatenate` / `div_tensor` and `scatter_reduce` / `scatter_reduce_` now compile and pass. The 5 failing CI checks are all on other vendor backends (tsingmicro / metax / mthreads-qwen / enflame-gcu400 / iluvatar-qwen) and are unrelated to these AMD-only changes; `amd36x-unit-test` is green.

zhzhcookie (approved):

